PR #1, opened 2017-03-17
New ensure user

omad:
I've rewritten the datacube-ensure-user script into python, and added the ability for it to copy user credentials in `~/.pgpass` for use with the new database server.

I've tested all the functionality myself, but it would be good to get another pair of eyeballs, and test by actual users, before we inflict this on everyone.

### DIFF
--- a/modules/agdc-instances/datacube-ensure-user.py
+++ b/modules/agdc-instances/datacube-ensure-user.py
@@ -49,12 +49,12 @@ def can_connect(dbcreds):
         return False
 
 
-def find_credentials(pgpass, old_host=OLD_DB_HOST):
+def find_credentials(pgpass, host):
     with pgpass.open() as src:
         filedata = [line.strip() for line in src]
 
     existing_db = [line for line in filedata
-                   if old_host in line][0]
+                   if host in line][0]
 
     return DBCreds(*existing_db.split(':'))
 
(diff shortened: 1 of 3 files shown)


modules/agdc-instances/datacube-ensure-user.py
```diff
@@ -0,0 +1,185 @@
...
+    try:
+        conn = psycopg2.connect(host=dbcreds.host, port=dbcreds.port, user=dbcreds.username)
+        cur = conn.cursor()
+        cur.execute('SELECT 1;')
+        return True
+    except psycopg2.Error:
+        return False
+
+
+def find_credentials(pgpass, host):
```
jeremyh: It's worth adding user as a parameter too (defaulting to current user name?), as it can have multiple entries for different roles on the same host.

(I have separate admin, backup, read-only users configured in my pgpass for the hosts)

modules/agdc-instances/datacube-ensure-user.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env python
```
jeremyh, line 1: Fantastic to see that this is Python now rather than bash 🥇

modules/agdc-instances/datacube-ensure-user.py
```diff
@@ -0,0 +1,185 @@
...
+    except FileNotFoundError:
+        filedata = []
+
+    new_creds_line = ':'.join(tuple(dbcreds))
+    filedata.append(new_creds_line)
+
+    os.umask(0o077)
+    with pgpass.open('w') as dest:
+        for line in filedata:
+            dest.write(line + '\n')
```
jeremyh: We should do this as an atomic file write just in case.

jeremyh (commented):
Looks good

PR #3, opened 2017-05-24
Changed FileNotFoundError to IOError

Kirill888:
This is to support py2. Currently failing to save credentials on agdc-py2-prod.

**!!!Not sure how to test this!!!**


omad:
I'm going to be a bit wild and crazy, and since we aren't already testing this very well, continue along in the same fashion.

PR #11, opened 2017-07-28
Documentation for creating/testing new product

Kirill888:
Adding new section that describes steps for testing new product using test database.

Also includes `pq_count_albers` product description.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -45,6 +45,7 @@ Publicly available data access and web services are currently in development.
    query/Getting Started.ipynb
    query/List Products.ipynb
    query/Loading Data.ipynb
+   internal/new_product.rst
 
 .. toctree::
    :maxdepth: 1
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -45,6 +45,11 @@ Publicly available data access and web services are currently in development.
    query/Getting Started.ipynb
    query/List Products.ipynb
    query/Loading Data.ipynb
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Developer Guide
+
    internal/new_product.rst
 
 .. toctree::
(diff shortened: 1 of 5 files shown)


docs/index.rst
```diff
@@ -45,6 +45,7 @@ Publicly available data access and web services are currently in development.
    query/Getting Started.ipynb
    query/List Products.ipynb
    query/Loading Data.ipynb
+   internal/new_product.rst
```
andrewdhicks: Start a new section (eg Developer Guide) to keep this level of detail away from the User Guide.

docs/internal/new_product.rst
```diff
@@ -0,0 +1,333 @@
...
+
+It's best to create an empty database for testing new product. Test database
+should contain same metadata and products as the main database. Typically new
+product is going to be derived from existing products in the main database and
+since we track dataset lineage we need to have the same products in the test
+database as in the main.
+
+First create empty database
+
+.. code-block:: bash
```
andrewdhicks: You're duplicating code from the appendix `using datacube_helpers.sh`
You could `literalinclude` with `:lines: ` to include subsets.
http://www.sphinx-doc.org/en/stable/markup/code.html

Kirill888: Yeah the problem is that I add extra line that runs the declared function, and Appendix is meant to be sourced without side-effects. Is there a way to combine `literalinclude` with extra text added?

Kirill888: done it via multiline include

omad (approved):
Looks great, useful information. Some things could be cleaned up and reworded, but for internal development docs we can get to that later.

Merging now to minimise the number of branches we have lying around everywhere.

PR #15, opened 2017-08-28
Find gaps in correspondence between products

uchchwhash:
Add a script `gaps.py` to the `digitalearthau/scripts` folder. It
provides a CLI interface to the function `find_gaps` and dumps
the generated summary in `yaml` to `stdout` (or optionally a file).
A 'gap' here arises when a dataset in one of the products does not
have a corresponding dataset having the same spatio-temporal domain
in the other product. The user should specify products to compare
by name, and may also specify a date range for the datasets.

### DIFF
--- a/scripts/gaps.py
+++ b/scripts/gaps.py
@@ -52,6 +52,10 @@ class Dataset:
         """ String representation by ID. """
         return str(self)
 
+    def to_dict(self):
+        """ `dict` representation. """
+        return dict(id=self.id_, **self.keys)
+
 
 class Tally:
     """ Tally of mismatches. """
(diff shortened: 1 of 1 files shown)


scripts/gaps.py
```diff
@@ -0,0 +1,310 @@
...
+              help="output .yaml file to write report to (default: stdout)")
+@click.option('--start-date', type=str,
+              help="start of date range (YYYY, YYYY-MM, or YYYY-MM-DD)")
+@click.option('--end-date', type=str,
+              help="end of date range (YYYY, YYYY-MM, or YYYY-MM-DD)")
+@click.option('--time-divs', type=int,
+              help="divide the date range (computational)")
+def main(products, output_file, start_date, end_date, time_divs):
+    """ Entry point. """
+    datacube = Datacube(app='find-those-gaps')
```
GypsyBojangles, line 311: Could we display usage information if no products are supplied?
:o)

uchchwhash: done?

PR #18, opened 2017-09-14
More packages, local packages

benjimin:
The benefits of the environment modules (maintaining everybody on the same versions) are more fully realised if fewer users need opt-out (due to lacking various components that are important to specific individuals).

This change aims to make it easier for users to locally test additional packages via `pip install --user`. (Also, bundles a computer vision library.)


modules/py-environment/modulefile.template
```diff
@@ -22,7 +22,7 @@ if {[info exists env(PYTHONPATH)] && ${esc}env(PYTHONPATH) != ""} {
 }
 
 # This disables loading modules and .pth files from ~/.local/lib/python2.7/...
```
jeremyh: It's worth updating this comment too

jeremyh (changes requested):
Thanks Ben, this looks like a great solution

modules/py-environment/modulefile.template
benjimin: I thought that comment was still correct; how would you like it updated?

modules/README.md
```diff
@@ -60,7 +60,7 @@ This will create a new environment module in `/g/data/v10/public/modules/agdc-py
 
 Loading the module might conflict with other python modules you have loaded.
 
-The module will disable locally installed python packages to prevent conflicts by setting `PYTHONNOUSERSITE`
+The module will prevent conflicts with locally installed python packages by changing `PYTHONUSERBASE` for each release.
```
uchchwhash: Probably mention what it is being changed to?
Do we need to add it to PYTHONPATH too? In my setup I add to PYTHONPATH the equivalent of `~/.digitalearthau/${version}/local/lib/python3.6/site-packages`. And there's also the matter of adding `~/.digitalearthau/${version}/local/bin` to PATH.
And thanks Ben for this!

modules/py-environment/modulefile.template
jeremyh: It would be more useful to say why it's being set. The old one didn't do that very well either, only explaining what the `PYTHONNOUSERSITE` variable was for.

Perhaps "# Make user packages folder unique to each module to avoid dependency conflicts", or similar?

modules/README.md
benjimin: No, don't need to explicitly add it to PYTHONPATH (instead the appropriate sub-directory will already appear in `sys.path` via a different mechanism), nor explicitly create the directories (pip handles that when necessary).

You're right about adding the `bin` to PATH.

Was initially inclined to keep the technical details out of the README, but perhaps explaining ("`pip install --user ...` will store things under `~/.digitalearthau`" is worthwhile..

jeremyh (approved):

PR #23, opened 2017-09-21
Add PBS qsub self submission lib

Kirill888:
This is cleaned up copy from datacube stats with tests added.

### DIFF
--- a/digitalearthau/qsub.py
+++ b/digitalearthau/qsub.py
@@ -1,0 +1,636 @@
+import click
+import yaml
+import sys
+import re
+import logging
+import itertools
+from time import sleep
+
+from pydash import pick
+from pathlib import Path
+from functools import update_wrapper
+from subprocess import Popen, PIPE
+
+from . import pbs
+from datacube.executor import (SerialExecutor,
+                               mk_celery_executor,
+                               _get_concurrent_executor,
+                               _get_distributed_executor)
+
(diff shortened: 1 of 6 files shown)


digitalearthau/qsub.py
```diff
@@ -0,0 +1,636 @@
...
+    """
+
+    py_file = str(Path(sys.argv[0]).absolute())
+    return (sys.executable, py_file) + args
+
+
+def generate_self_launch_script(*args):
+    s = "#!/bin/bash\n\n"
+    s += pbs.generate_env_header()
+    s += '\n\nexec ' + ' '.join("'{}'".format(s) for s in self_launch_args(*args))
```
jeremyh: There's a handy built-in for quoting shell args (that'll handle other special characters too): `shlex.quote(arg)`

scripts/qsub_sample_app.py
```diff
@@ -0,0 +1,108 @@
...
+
+def task_generator(num_tasks):
+    """ generates sample tasks
+    """
+    for i in range(num_tasks):
+        click.echo('Generating task: {}'.format(i))
+        yield Task(i)._asdict()
+
+
+# pylint: disable=invalid-name
```
jeremyh: Why did this need to be disabled? Might be worth adding a comment above it

jeremyh (approved):
Thanks Kirill

PR #26, opened 2017-09-27
Add `dea-system init` command

jeremyh:
This adds a command similar to the `datacube system init` command, but that will create the system with DEA products and configuration.

```
jez@kveikur:~/prog/ga-datacube-env (develop) $ dea-system init 
ODC init of postgresql://jez@:5432/dev
    Created.
    Checking indexes/views.
Checking DEA metadata types
    eo
    telemetry
    landsat_scene
    landsat_l1_scene
Checking DEA products
    low_tide_comp_count
    high_tide_comp_count
    ls8_satellite_telemetry_data
    ls8_level1_scene
    ls8_nbar_scene
    ls8_nbart_scene
    ls8_pq_scene
    ls8_level1_oli_scene
    ls8_nbar_oli_scene
    ls8_nbart_oli_scene
    ls8_pq_oli_scene
    ls7_satellite_telemetry_data
    ls7_level1_scene
    ls7_nbar_scene
    ls7_nbart_scene
    ls7_pq_scene
    dsm1sv10
    ls5_satellite_telemetry_data
    ls5_level1_scene
    ls5_nbar_scene
    ls5_nbart_scene
    ls5_pq_scene
    pq_count_albers
    high_tide_comp_20p
    low_tide_comp_20p
Checking DEA ingested definitions
    ls8_pq_albers       		← ls8_pq_scene
    ls8_nbar_albers     		← ls8_nbar_scene
    ls8_nbart_albers    		← ls8_nbart_scene
    ls7_nbar_albers     		← ls7_nbar_scene
    ls5_nbart_albers    		← ls5_nbart_scene
    ls7_nbart_albers    		← ls7_nbart_scene
    ls8_nbart_oli_albers		← ls8_nbart_oli_scene
    ls7_pq_albers       		← ls7_pq_scene
    ls8_nbar_oli_albers 		← ls8_nbar_oli_scene
    ls5_nbar_albers     		← ls5_nbar_scene
    ls8_pq_oli_albers   		← ls8_pq_oli_scene
    ls5_pq_albers       		← ls5_pq_scene
jez@kveikur:~/prog/ga-datacube-env (develop) $ 
```

Unlike the manual process used previously, this creates products for ingested datasets too.


digitalearthau/duplicates.py
```diff
@@ -140,26 +141,28 @@ def _write_csv(unique_fields, dicts, stream, append=False):
...
-            'Or specify --all to check all\n'.format(
-                sys.argv[0], '\n\t'.join(sorted(collections.registered_collection_names())))
-        )
-        sys.exit(1)
+collections.init_nci_collections(None)
+
+
+@click.command('duplicates')
+@global_cli_options
+@click.option('-a', '--all_', is_flag=True)
```
Kirill888, line 149: what's up with postfix underscores?

Kirill888 (approved):

andrewdhicks (approved):
Looks good! 
I can see this leading to a test comparing what is in the database vs what is in the config files.

digitalearthau/duplicates.py
jeremyh: The argument to that method is `all_` to avoid shadowing the `all` built-in. It probably shouldn't be exposed on the cli, though. The convention comes from [pep8](https://www.python.org/dev/peps/pep-0008/#id46): 

> If a function argument's name clashes with a reserved keyword, it is generally better to append a single trailing underscore rather than use an abbreviation or spelling corruption. Thus class_ is better than clss. (Perhaps better is to avoid such clashes by using a synonym.)

Kirill888:
```python
@click.option('-a', '--all', 'all_', is_flag=True)
```

will give `--all` command line option and will store it into `all_` variable

PR #27, opened 2017-09-28
Update pq count product yaml

Kirill888:
- uint16 -> int16
- nodata -> -1
- product_type  -> pq_count
- Split annual vs all time product


digitalearthau/config/products/pq_count_annual_summary.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+metadata:
+  product_type: pq_count
+  stats:
+    period: '12m'
+  format:
+    name: NetCDF
+
+storage:
+  driver: NetCDFCF
```
jeremyh: It should be 'NetCDF CF' with a space. We should add something to validate this...

jeremyh: ... and after checking, it looks like some of the other new products have the same mistake too. I'll create a separate issue for those.

```
$ rg 'driver: '

digitalearthau/config/ingestion/ls8_pq_albers.yaml
94:  driver: NetCDF CF

digitalearthau/config/ingestion/ls8_nbar_albers.yaml
109:  driver: NetCDF CF

digitalearthau/config/ingestion/ls8_nbart_albers.yaml
115:  driver: NetCDF CF

digitalearthau/config/ingestion/ls7_nbar_albers.yaml
109:  driver: NetCDF CF

digitalearthau/config/products/pq_stats.yaml
12:  driver: NetCDFCF

digitalearthau/config/ingestion/ls5_nbart_albers.yaml
115:  driver: NetCDF CF

digitalearthau/config/ingestion/ls7_nbart_albers.yaml
115:  driver: NetCDF CF

digitalearthau/config/ingestion/ls8_nbart_oli_albers.yaml
115:  driver: NetCDF CF

digitalearthau/config/ingestion/ls7_pq_albers.yaml
94:  driver: NetCDF CF

digitalearthau/config/ingestion/ls8_nbar_oli_albers.yaml
109:  driver: NetCDF CF

digitalearthau/config/ingestion/ls5_nbar_albers.yaml
109:  driver: NetCDF CF

digitalearthau/config/ingestion/ls8_pq_oli_albers.yaml
94:  driver: NetCDF CF

digitalearthau/config/ingestion/ls5_pq_albers.yaml
94:  driver: NetCDF CF

digitalearthau/config/products/HTC_2_0_0.yaml
44:  driver: NetCDFCF

digitalearthau/config/products/LTC_2_0_0.yaml
44:  driver: NetCDFCF

digitalearthau/config/products/LTC_COUNT_2_0_0.yaml
24:  driver: NetCDFCF

digitalearthau/config/products/HTC_COUNT_2_0_0.yaml
24:  driver: NetCDFCF

```

digitalearthau/config/products/pq_count_annual_summary.yaml
```diff
@@ -0,0 +1,37 @@
+name: pq_count_annual_summary
+description: Pixel Quality Statistics 25m
+
+metadata_type: eo
+
+metadata:
+  product_type: pq_count
+  stats:
+    period: '12m'
```
jeremyh: I wonder if this may be better as a new common field such as `product_variant`, as non-stats products might have variations too? The purpose of it, as I understand, is to distinguish the variations of the `product_type` rather than actually give a precise period (eg. there may not be a whole year of data), especially in the other variants, like `'*'` ("full") below, or seasonal. The precise period is recorded in the extent start/end times.

If we *do* want to distinguish the period specifically, maybe `extent → selection_period`, or something? but that would need precise definitions of the possible values, unlike the arbitrary name allowed by a `product_variation` field.

Using a common field would also make it easier to show different product variations together on a UI without needing to know about different specific fields per product.

digitalearthau/config/products/pq_count_annual_summary.yaml
andrewdhicks: Those new products can be fixed under PR #28

digitalearthau/config/products/pq_count_annual_summary.yaml
andrewdhicks: I've had a similar issue with hltc and item:
```
 +    product_type: tidal_composite
 +    statistic:
 +        name: clearcount
 +        source: pq
 +        tidal_range:
 +            name: high
 +            min: 80
 +            max: 100
```

digitalearthau/config/products/pq_count_annual_summary.yaml
Kirill888: Also pretty much all stats products have missing space

digitalearthau/config/products/pq_count_annual_summary.yaml
Kirill888: I think whole `product_type` concept is misleading, and it doesn't help that it is not backed by any kind of table listing all valid `product_type`s. I think one should be able to just add arbitrary number of tags to any product, because that's what product type currently is: a tag, except only one tag is allowed.

Kirill888: The only reason I added that field is for auto-match to work, it has no value for describing product or anything like that. For stats products period of accumulation is in the product name, but I can't add datasets by product name only by product_type + whatever other metadata.

digitalearthau/config/products/pq_count_annual_summary.yaml
andrewdhicks: I thought the `driver` part of `storage` was being phased out, as it isn't a property of the dataset, only the storage mechanism.

Kirill888: What are the situations where `storage` subtree is used?

I just copied stuff from some other product definition without any understanding. What code currently cares about that section of product definition?

Cause I have been able index datasets with that product description and then query them through datacude and load data without any problems what so ever.

omad: The `storage` section is critical whenever we are writing data. I can't remember (it's not clearly delineated) whether it's ever referred to when reading data.

In most cases we trust what's in the file itself. But there are cases where it can be overridden.

digitalearthau/config/products/pq_count_annual_summary.yaml
Kirill888: We discussed this in person, but I think it's worth recording here also. I don't think `product_variant` is a good idea.  In my understanding all products that share the same product type are already variants, and what distinguishes them is context specific. In the case of `nbar` differences come from the sensor/platform, in the case of statistic products it is usually period of aggregation. In some other context it might very well be some `variant` field containing a unique string (product name maybe?). Do we need to make that a "first class entity" as far as code and database is concerned? Probably not.

PR #28, opened 2017-10-03
ITEM and HLTC product configs

andrewdhicks:
Adding Intertidal Extent Model and High/Low Tide Composite product configs.
(Starting PR early for visibility of product_type issues and driver issues)


omad (approved):
Looks good to me. We can improve when there are more products in the future, if necessary.

PR #32, opened 2017-10-11
Add events, logging handling and task state tracking

jeremyh:
With a celery implementation.

(this requires the opendatacube/datacube-core#310 patch for "useful" task state events, but the logging, work directory management and subjob handling will work without it)

Example output folders for a `task_app` process (`fc`), written to `/g/data/v10/work` by default:
```
.
`-- ls8_fc_albers
    `-- create
        `-- 2017-10
            `-- 12-024657
                |-- events
                |   `-- 1507737024-r301-collected-events.jsonl
                |-- jobs
                |   |-- 1507736817-generate-9688372.r-man2.yaml
                |   `-- 1507736874-run-9688415.r-man2.yaml
                |-- logs
                |   |-- 1507736817-generate-head.err.log
                |   `-- 1507736817-generate-head.out.log
                |   |-- 1507736874-run-head.err.log
                |   `-- 1507736874-run-head.out.log
                |-- generated-tasks.pickle
                `-- task-description.json
```
(the "1507737024-" prefix is an epoch timestamp similar to the version field on our *.nc files, primarily so they're ordered by date)

Note that the folder structure differs slightly from our previous discussion: grouping information by the output product, rather than the software (I'm happy to change this if someone objects?):
```
/work/ls8_fc_albers/create/2017-10/12-024657
/work/{output_product}/{task_type}/YYYYMM/DD-HHMMSS
```
- `create` in this case can also be `sync`, `archive` or whatever other thing we may do to the product  

Here's an example `fc` patch to show an app using these apis: GeoscienceAustralia/fc#10


andrewdhicks (commented):
Looks good to me, although I'm not across the celery portion.

Kirill888:
It's a lot of changes, but I looked through qsub/celery sections, changes there look fine. Thanks for adding all the docs and type annotations 👍.

andrewdhicks (approved):

digitalearthau/runners/model.py
```diff
@@ -0,0 +1,71 @@
...
+
+    pbs_parameters: PbsParameters = None
+
+
+class DefaultJobParameters(NamedTuple):
+    # Datacube query args used to select datasets to process (eg time=(1994, 1995), product=ls5_nbar_albers)
+    query: dict
+
+    # Input and output products ("ls7_nbar_albers")
+    source_types: List[str]
```
omad, line 46: Could these be named `source_products` and `output_products`. Types is such an overloaded name...

digitalearthau/runners/model.py
```diff
@@ -0,0 +1,71 @@
...
+    output_types: List[str]
+
+
+class TaskDescription(NamedTuple):
+    # task type (eg. "fc")
```
omad, line 51: Would a better example be "create" or "sync" or "stack"?

digitalearthau/runners/util.py
```diff
@@ -0,0 +1,151 @@
...
+from digitalearthau import pbs
+
+from ..qsub import QSubLauncher, with_qsub_runner, norm_qsub_params, TaskRunner
+from .model import TaskDescription, DefaultJobParameters, TaskAppState, PbsParameters
+from .. import serialise, paths
+
+_LOG = logging.getLogger(__name__)
+
+
+def init_task_app(
```
omad, line 22: This seems like it should be a standard function for using `task_app`, rather than merely a "convenience" function hidden in the `utils` module...

If we're doubling down on `task_app` I think it would make more sense to move this and add a template/documentation on how it should be used.

digitalearthau/runners/util.py
```diff
@@ -0,0 +1,151 @@
...
+    return task_desc, task_desc_path
+
+
+def submit_subjob(
+        name: str,
+        task_desc: TaskDescription,
+        command: List[str],
+        qsub_params) -> str:
+    """
+    Convenience method for submitting a sub job under the given task_desc
```
omad, line 80: Same comment as above on the "convenience" vs standardisation-ness of this function, if we want all our apps to be consistent.

digitalearthau/serialise.py
```diff
@@ -0,0 +1,238 @@
...
+import datetime
+import dateutil.parser
+import yaml
+from typing import Dict
+
+from digitalearthau import paths
+from digitalearthau.events import Status
+
+
+class JsonLinesWriter:
```
omad, line 15: Although the name is somewhat self explanatory, I think it would be good to have a docstring describing that this is for streamed writing of a JSON log file. Presumably to be scraped up by ElasticSearch or some other log aggregator?

digitalearthau/serialise.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+
+    # We can't isinstance() for NamedTuple, the noraml way is to see if attributes like _fields exist.
+    try:
+        return dict(zip(o._fields, (type_to_dict(value) for value in o)))
+    except AttributeError:
+        return simplify_obj(o)
+
+
+def dict_to_type(o, expected_type):
```
omad, line 166: This is cool, it's pretty neat using type annotations for parsing/validating data, but it seems like it could easily grow out of control and be yet more work to maintain/document/support...

digitalearthau/uiutil.py
```diff
@@ -24,50 +26,10 @@ def init_logging():
             structlog.processors.StackInfoRenderer(),
```
omad, line 26: Just a question: Should this `init_logging()` function be always called from `task_app` initialisation? In the interest of standardising things as much as possible.

digitalearthau/serialise.py
```diff
@@ -0,0 +1,238 @@
+import enum
```
omad, line 1: Am I right in my reading that this module contains basically two useful bits of functionality:
- A class for writing out to JSON-Lines type files
- Functions for serialising/deserialising objects based on a specifications provided by a subclass of NamedTuple, which includes type annotations on each field
?

It's the second part I haven't completely grokked yet. :)

digitalearthau/runners/util.py
omad: Another option would be moving this functionality into the `TaskDescription` itself, although the name may not make as much sense...

digitalearthau/runners/celery_environment.py
```diff
@@ -0,0 +1,332 @@
...
+from datacube import _celery_runner as cr
+from digitalearthau import serialise, pbs
+
+from digitalearthau.events import Status, TaskEvent, NodeMessage
+from digitalearthau.runners.model import TaskDescription
+
+_LOG = logging.getLogger(__name__)
+
+# The strigified args that celery gives us back within task messages
+_EXAMPLE_TASK_ARGS = "'(functools.partial(<function do_fc_task at 0x7f47e7aad598>, {" \
```
omad, line 30: These two constants don't appear to be used. Are they just here as example documentation? (which does seem useful). In which case as a nicely formatted comment would be a better format.

digitalearthau/pbs.py
```diff
@@ -46,6 +51,44 @@ def load_lines(fname):
     return [Node(**x) for x in _nodes.values()]
 
 
+# This is defined in document "DEA Event structure", to produce stable & consistent task ids for pbs jobs.
```
omad, line 54: Would be great to provide a link to [DEA Event structure](https://docs.google.com/document/d/1VNpK3GL1r4kbjwAO-sJ6_BMk2FSHhNnoDg4VHeylyAE/edit?usp=sharing) or include it in the project docs.
I've changed it to only be viewable by people with the link, not editable.

digitalearthau/paths.py
```diff
@@ -299,3 +318,37 @@ def trash_uri(uri: str, dry_run=False, log=_LOG):
         if not trash_path.parent.exists():
             os.makedirs(str(trash_path.parent))
         os.rename(str(base_path), str(trash_path))
+
+
+def get_product_work_directory(
+        output_product: str,
+        time=datetime.datetime.utcnow(),
```
omad, line 325: At the moment I'd only expect this function to be called once.

But it's probably a gotcha in the future if anyone tried to call it a few times from the some process and they were given identical timestamps.

digitalearthau/events.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class NodeMessage(NamedTuple):
+    hostname: str
+    pid: int
+
+    # An optional id to identify a single instance within a host, such as a worker. (pid is not unique)
+    runtime_id: Optional[uuid.UUID] = None
+
+    def current_node(self) -> 'NodeMessage':
```
omad, line 19: Should be a `@classmethod`? Helps to mark it as an alternative constructor.

Or is it supposed to be called from an instance?

digitalearthau/serialise.py
omad: Also, I've just worked out that it will be used for logging `TaskEvents`. Sure it's flexible, but having context is good too.

omad (approved):
The code looks great. But the lack of high level documentation made it take quite a while to understand what was going on.

I'd love to see more explanatory documentation:
- Module level docstrings, overview of the important classes/methods, what they're useful for or where they're being used.

- Documentation explaining the concept of a **task**, made up of potentially multiple PBS jobs, log files, other artefacts, that will be linked together by an id an other info in a `TaskDescription`, which lives in python land as an object, and is serialised out to the **task directory**. It's the responsibility of each execution to load the `TaskDescription` from disk to know what it's a part of. Or is it worked out by parameters passed between jobs? Or by working directory? I'm not sure...

There's also no explanation linking `TaskEvent` to the logging that's happening in the Celery executor, using the `JSONLinesWriter`, or how that's intended to be used.

I'm going to approve and merge to keep things moving, since as I said, the code looks great. I'm happy to help with some high level documentation when there's time. :)

(famous last words, I know :) )

digitalearthau/events.py
jeremyh: Yep, it should. It's not used anymore in the current code so I may be better to remove it entirely.

digitalearthau/paths.py
jeremyh: Good catch. It'll fail in subsequent calls because the directory already exists, but that's not an ideal failure case. It'll also fail for the same reason if called multiple times within the same second for the same type of product/process (due to directory name granularity of HHMMSS). 

It's good enough for our current cli-usage of creating infrequent large jobs (and failures will be loud, not silent), so I don't think it's a blocker, but will need some improvement for wider use cases.

digitalearthau/runners/celery_environment.py
jeremyh: The latter is used in the doctests to test extracting data from such strings. The former isn't currently used, but may be if we need more data from it in the future.

digitalearthau/runners/model.py
jeremyh: Yep, good idea. The names were originally mirroring the fields in `task_app` config files, but your suggestion sounds clearer. We should change this now if we ever want to change it.

digitalearthau/runners/model.py
jeremyh: Possibly. At the very least, the inconsistency with the newer work folder structure is strange. I'll try to define them better in the events spec...

digitalearthau/runners/util.py
jeremyh: Yeah, agreed. These functions were moved here from FC at the last minute after some cleanup there. 

I think of these (and the current task_app) as too lower level, and not the eventual API we want products to be calling. I'd like to move the whole CLI interface to DEA, with products themselves only implementing run/query functions.

digitalearthau/serialise.py
jeremyh: Sorta. The latter were written to output NamedTuples as dicts. By default json and other serialization modules output them as lists of values (ie: as tuples). I'll improve the comments.

The use of named tuples were an attempt to be a compromise between dicts and full classes. I A lot of the previous dict-based code seemed to grow unwieldy, especially as it neutered a lot of our tools (pycharm/pylint). Normal classes are an option too, but they'd need their own serialization code anyway.

digitalearthau/uiutil.py
jeremyh: It's used at the moment by the other dea tools (sync etc). In task_app, logging has to be handled by the runner as they all have creative ways of spawning/forking/creating/pbsdsh'ing separate processes, and for now I backed-out those changes due to their fragility.

PR #37, opened 2017-10-19
Dask-worker memory limit 'auto' to 4GB

harshurampur:
Increasing memory limit from 'auto' to 4gb encountering the error message below:

```
distributed.worker - WARNING - Memory use is high but worker has no data to store to disk.  Perhaps some other process is leaking memory?  Process memory: 791.19 MB -- Worker memory limit: 1.05 GB
```


andrewdhicks (approved):
Great!

PR #38, opened 2017-10-19
 Preliminary NCI test environment

uchchwhash:
Tools to maintain a more persistent test database
than those created for integration tests.

- Add command `test-env setup` and `test-env teardown`
  that sets up and tears down the test database respectively
- Add command `test-env migrate` that indexes datasets
  from the production database that match the given query
  into the test database
- Add test for `metadata_type` inconsistency in ingestion


digitalearthau/test_env.py
```diff
@@ -0,0 +1,354 @@
...
+# it's used here to be backwards compatible.
+default_environment: datacube
+
+[datacube]
+# Inherit all defaults.
+"""
+
+    class LocalConfig(ConfigBase):
+        """
+        Backporting the environment feature in `develop` to `release-1.5`.
```
jeremyh, line 87: Audacious 🙂 

It's probably useful enough that it might be worth actually putting in release-1.5.

digitalearthau/config/datacube.conf
```diff
@@ -0,0 +1,9 @@
+[datacube]
+db_hostname: agdc-db
+db_port: 6432
+db_database: datacube
+
+[NCI-test]
```
jeremyh: These commands look useful enough that individual developers would probably run them for their own cubes.  Perhaps we could default the `db_database` to the current username, or something unique, so that they can run it out-of the-box without risk of conflicting? (current_username is typical postgres convention for a user's default database)

digitalearthau/test_env.py
```diff
@@ -0,0 +1,354 @@
...
+def psql_command(command, local_config, maintenance_db='postgres'):
+    """
+    Feed ``command`` to the PostgreSQL server specified in ``local_config``.
+    """
+    hostname = local_config.db_hostname
+    port = local_config.db_port
+
+    # seems like you have to connect to a database
+    # and that the maintenance database is usually called postgres
+    return run_shell(["psql", "-h", hostname, "-p", port, maintenance_db],
```
jeremyh, line 163: If this script keeps growing it would be worth using one of the Python Postgres apis rather than calling to a shell command. Such as the Index object's own connection, or plain psycopg2. 

The commands calling this are manually formatting SQL strings, but it would be better to use something that escapes arguments for us to avoid issues with special characters. This is pretty insignificant at the moment (all arguments come from config files, and we're not using special characters in our database names), but everything has a habit of growing beyond its original usage 🙂

digitalearthau/test_env.py
```diff
@@ -0,0 +1,354 @@
...
+    polygon = query.geopolygon
+    if polygon:
+        for dataset in datasets:
+            if intersects(polygon.to_crs(dataset.crs), dataset.extent):
+                yield dataset
+    else:
+        yield from datasets
+
+
+def normalize_uri(uri):
```
jeremyh, line 261: The property `dataset.local_path` could simplify these two methods — the indexing function we use here doesn't currently support uris anyway, and you only need the latest path, I think? We've got too many uri-to-path implementations already 😛

integration_tests/test_config.py
```diff
@@ -66,3 +66,14 @@ def test_dea_config(dea_index: Index):
         'pq_count_summary',
         'pq_count_annual_summary',
     }
+
+
+def test_metadata_type(dea_index):
```
jeremyh, line 74: 👍

setup.py
```diff
@@ -59,6 +59,7 @@
             'dea-submit-sync = digitalearthau.sync.submit_job:main',
             'dea-sync = digitalearthau.sync:cli',
             'dea-system = digitalearthau.system:cli',
+            'test-env = digitalearthau.test_env:cli'
```
jeremyh: This command gets added to everyone's global path, so we don't want command names to be too generic (and possibly conflict with others). 

There's also nothing specifically test-related about it: it's creating a real datacube from another. Perhaps call it `dea-env`? (renaming this module and the above application_name too)

```
   dea-env setup
   dea-env migrate
   dea-env teardown
```


To take the idea further, and don't feel pressured to do this in this pull request, `setup` could be moved into `dea-system`, which currently has an `init` command, but could explicitly create them too.

Something like:
```
 dea-system create
 dea-import  --from-env prod <search_args>
 dea-system drop
```

digitalearthau/test_env.py
```diff
@@ -0,0 +1,354 @@
...
+    # for the URIs collected, copy them to a local file system
+    # and change the URIs accordingly before indexing in the next section
+    # the digitalearthau.move module seems relevant
+    # it would be great if the user can provide a mapping of file systems
+    # perhaps a .yaml file with multiple entries like
+    # - source: /g/data/rs0/datacube/
+    #   target: /g/data/u46/users/ia1511/data/datacube/
+    # or something similar
+
+    # there are too many DuplicateRecordError warnings
```
jeremyh, line 299: Those warnings bug me too; it's normal behaviour to add a dataset with duplicate ancestor datasets of another. We should add an issue.

jeremyh (commented):
Looks good Imam. I'm slightly hesitant about the amount of duplicated functionality from elsewhere, as it's more to maintain in the future (ideally we should improve the upstream methods 🙂 ), but it's not a blocker in the short term.

uchchwhash:
Notwithstanding the fact that I have not worked through all of @jeremyh's suggestions here, I don't think it would be desirable to merge this branch as it is here. Instead, I am starting anew at the `NCI-test-env` branch that targets core `develop` branch soon to be released as release 1.5.6.

PR #39, opened 2017-10-20
Expand module docs for serialise, events

jeremyh:
This is only a start. As suggested by Damien in #32


omad (approved):
Brilliant, helps pull it together. Thanks Jeremy.

PR #42, opened 2017-10-26
Store cleanup results consistently & automatically

jeremyh:
The collection management scripts currently output their results to stdout, leaving it up to the user
to pipe/record the results somewhere.

But they are typically run in an ad-hoc manner by different users, so this previous output has mostly been lost or is disorganised.

These commits change the script to automatically output results to the common work folder structure (as with the new `fc`) without user intervention, and to print readable output to the user (no json) about the status.

(this request only changes `cleanup`, but I expect all of the management scripts to eventually do this too)

### DIFF
--- a/digitalearthau/cleanup.py
+++ b/digitalearthau/cleanup.py
@@ -67,14 +67,24 @@ def indexed(index: Index,
 
     for product, datasets in index.datasets.search_by_product(**expressions):
         expected_count = product_counts.get(product.name, 0)
+
+        # Record results to the product's work folder.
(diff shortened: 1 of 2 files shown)


Kirill888 (approved):
I'm not familiar with the original code, so... but changes look reasonable to me, with good comments and docs. 

I noticed you are using format strings, is this a new addition or is this repo 3.6+ only already?

jeremyh:
We made this and other GA-only repos py-3.6-only as that's what our own environment is using, and is the only version we test against.

PR #43, opened 2017-10-27
Clean up sync tests, add move tests

jeremyh:
There's one move test marked as xfail: it tests whether you can move a dataset when the destination already exists. But this is no longer likely to be implemented this week, so I've left it disabled. The current behavior is to skip those datasets, logging a warning.

### DIFF
--- a/digitalearthau/cleanup.py
+++ b/digitalearthau/cleanup.py
@@ -1,6 +1,3 @@
-"""
-Find and trash archived locations of active datasets.
-"""
 from datetime import datetime, timedelta
 from pathlib import Path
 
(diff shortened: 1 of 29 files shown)


jeremyh:
I've just pushed two more commits that greatly improve the click help information on all of the collection tools, and adds a documentation page introduction to them. 

They're unrelated enough that they probably should be a separate pull request, but they're currently dependent on some of the changes in this pull request.

omad (approved):
Crikey. I've been starting at this all morning, I think it all looks good, but there's a lot of changes!

PR #45, opened 2017-10-29
October miscellaneous

omad:
- Don't install `agdc-statistics` with `datacube-core`. We're going to keep it in a separate module now, so that when we want a new release of it we don't need a new release of anything else.
- I don't recall why pylint was pinned to 1.6.4 everywhere, but that release is broken for me...
- Ignore mypy cache directory, to hopefully start running static type checks across the codebase with mypy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ docs/html/
 
 # Generated Documentation
 docs/dev/generate/
+
+# MYPY static type checker cache
+.mypy_cache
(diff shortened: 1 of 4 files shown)


jeremyh:
As discussed in Slack, we probably do want to pin pylint: new versions of pylint add new code checks, which cause our builds (and all tagged releases) to start failing. It's better to be in control of those version bumps. But it wasn't pinned with its dependencies (particularly astroid), causing it to break as new versions of astroid came out. 

We can re-pin it properly in another issue, for now it's worth unpinning.

PR #47, opened 2017-11-01
Enable no-else-return

Bis-Bala:
"no-else-return" is removed from disable list. The change is in configuration file pylintrc.


omad:
Thank's for re-enabling that pylint check!

However, looking at the other change in `dataset_prepare_prod.py`. It looks like **that script should be removed** instead of modified. As far as I know we're not using it, and I have a lot of issues with it:

- The concept of reading dataset metadata from a NetCDF then modifying it and writing out to disk is troublesome. We need to keep consistent metadata.
   1. Every effort should be made to write correct metadata into NetCDF files in the first place. 
   2. Failing that, the metadata inside the NetCDF files should be fixed, not just create a new metadata file in a new location.
- The new `ga-metadata.yaml` files don't contain any provenance history
- There's inadequate documentation on the purpose of the script
- The command line defaults look like they would create broken output, instead of prompting the user with an error.
- Copy/paste duplication within the code itself
- Not cleaning up file handles from opening NetCDF files

jeremyh:
I don't feel strongly about the issue, so feel free to continue, but I'd mildly prefer to disable `no-else-return`. I think symmetrical return statements can occasionally be more readable in small methods. It has the usual downsides of increased-nagging of pylint, as with all checks, but without adding enough benefit to be worth it.

Note that it's currently disabled in the datacube-core repository too (both by [myself](https://github.com/jeremyh/datacube-core/commit/1d227db84a52976708585f6e49f8c5e0d573e025) and [Damien](https://github.com/opendatacube/datacube-core/commit/24b3b0d2fed66f5710696e05e3ca9a4cb004586a) independently), and it frustrates developers less when the rules are kept consistent between repositories.

omad (approved):
Thanks!

PR #48, opened 2017-11-01
Add a DEA TaskApp version of Stacker

omad:
In an attempt to run all of our tasks in a consistent manner, this PR contains a new tool, `dea-stacker`. It is able to stack individual time slice datasets into a time-deep _stacked NetCDF_ file format. It is hard-coded to stack all available slices for a year into a single file. It operates on the same configuration file as `ingest` or `fc_app`, and so works on the same Grid Specification for tiling.

All the code it runs is pulled in from `datacube_apps.stacker`, but is wrapped in the typical **Submit**, **Generate**, **Run** commands used for running DEA processes on the NCI.

The app is based on `fc_app.py` in the [Fractional Cover Repo](GeoscienceAustralia/fc), which means a mass of duplication. My intention is to have a look at what is required, and replace both with a more extensible type of `TaskApp/DEATaskApp`, but this is work for the future.

### DIFF
--- a/digitalearthau/stacker.py
+++ b/digitalearthau/stacker.py
@@ -127,7 +127,7 @@ def make_tag(task_desc: TaskDescription):
 def generate(index: Index,
              task_desc_file: str,
              no_qsub: bool):
-    config, task_desc = _make_config_and_description(index, Path(task_desc_file), tag)
+    config, task_desc = _make_config_and_description(index, Path(task_desc_file))
 
     num_tasks_saved = task_app.save_tasks(
         config,
--- a/digitalearthau/stacker.py
+++ b/digitalearthau/stacker.py
@@ -3,22 +3,26 @@ A DEA version of `stacker` from ODC
 
 Uses the exact same implementation, but wraps it in QSUB Running/Event logging goodness.
 
-As such, it is a three step process:
+As such, it is a three step process, with the second two happening automatically after the first:
+
 1. Submit
-    Submit a single threaded PBS job to calculate how big the requested job is.
- - Executes very quickly
- - Creates the 'job' directory which groups logs and metadata about all the steps
- - Schedules a 'Generate' job to run through PBS
+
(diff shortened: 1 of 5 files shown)


digitalearthau/stacker.py
```diff
@@ -0,0 +1,242 @@
...
+# pylint: disable=invalid-name
+from digitalearthau.qsub import with_qsub_runner, TaskRunner
+from digitalearthau.runners.model import TaskDescription
+from digitalearthau.runners.util import init_task_app, submit_subjob
+
+_LOG = logging.getLogger(__file__)
+APP_NAME = 'dea-stacker'
+
+
+@click.group(help='DEA Stacker\n\n' + __doc__)
```
jeremyh, line 54: I never thought of using `__doc__` to deduplicate click help. I'm going to copy this in the other scripts :)

digitalearthau/stacker.py
```diff
@@ -0,0 +1,242 @@
...
+import datacube
+import digitalearthau
+from datacube.api.query import Query
+from datacube.index._api import Index
+from datacube.ui import click as ui
+from datacube.ui import task_app
+from datacube_apps.stacker import stacker
+from digitalearthau import __version__
+from digitalearthau import paths, serialise
+# pylint: disable=invalid-name
```
jeremyh, line 45: What name was this disabling? Still necessary?

jeremyh (approved):
👍

digitalearthau/stacker.py
omad: Thanks.
I'd never thought of it earlier either but it works really well. It's great having lots of information available on the command line.

digitalearthau/stacker.py
omad: Probably not, I'll disable it and see if anything complains.
Was probably for a click parameter that I stopped using.

PR #51, opened 2017-11-14
Start a common integration test module

jeremyh:
This allows the integration test config to be reused in other repositories.

To use it, one can add the following to `conftest.py`:

```python
    pytest_plugins = "digitalearthau.testing"
```

And then fixtures can be used, eg:

```python

def test_parse_query_args(dea_index: Index):
    """
    A user gives time start/end: they should be parsed as a single time field,
    and restricted to the current product.
    """

    product = dea_index.products.get_by_name('ls7_level1_scene')
    ... etc
```

(I'd like to add fixtures for initialising data too, but i'd prefer to clean up the current dea ones before exposing them "publicly")

### DIFF
--- a/digitalearthau/testing.py
+++ b/digitalearthau/testing.py
@@ -14,6 +14,9 @@ from datacube.index.postgres import PostgresDb
 from datacube.index.postgres import _dynamic
 from datacube.index.postgres.tables import _core
 
+# These are unavoidable in pytests due to fixtures
+# pylint: disable=redefined-outer-name,protected-access,invalid-name
+
 try:
     from yaml import CSafeLoader as SafeLoader
 except ImportError:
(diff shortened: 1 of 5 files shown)


jeremyh:
(the short-term use of this is for tests in the Dashboard repo)

digitalearthau/testing.py
```diff
@@ -0,0 +1,117 @@
...
+
+@contextmanager
+def _increase_logging(log, level=logging.WARN):
+    previous_level = log.getEffectiveLevel()
+    log.setLevel(level)
+    yield
+    log.setLevel(previous_level)
+
+
+@pytest.fixture()
```
omad: Is there an intended difference between `@pytest.fixture()` and `@pytest.fixture`?

omad (approved):

digitalearthau/testing.py
jeremyh: Nope; presumably a relic of when there were arguments on it.

PR #52, opened 2017-11-15
Improve dea-move to support NetCDF datasets

omad:
Add support to `dea-move` to move datasets comprising ODC NetCDF files. These files don't contain a separate checksum file, so it is necessary to run `dea-move` with the `--no-checksum` option.

I've also implemented a test for moving of these types of files, and made a few refactors to try and improve the code readability.

Also:
- Include new directories to `paths.py` and `collections.py` for Fractional Cover Dataset management
- Improve documentation on creating new Products on the NCI

### DIFF
--- a/digitalearthau/move.py
+++ b/digitalearthau/move.py
@@ -16,7 +16,9 @@ from eodatasets import verify
 from datacube.index._api import Index
 from datacube.model import Dataset
 from datacube.ui import click as ui
-from digitalearthau import collections, uiutil
+from digitalearthau.collections import init_nci_collections, get_collections_in_path
+from digitalearthau.paths import is_base_directory, BASE_DIRECTORIES, get_dataset_paths, split_path_from_base
+from digitalearthau.uiutil import init_logging
 from digitalearthau import paths as path_utils
 
 _LOG = structlog.get_logger()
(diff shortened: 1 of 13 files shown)


digitalearthau/move.py
```diff
@@ -200,39 +213,55 @@ def _do_copy(self, dry_run=True, checksum=True):
...
+        to_directory = to.parent
+        log.debug("copy.mkdir", dest=to_directory)
+        fileutils.mkdir_p(to.parent)
+        # We don't want to risk partially-copied files left on disk, so we copy to a tmp name
+        # then atomically rename into place.
+        tmp_name = tempfile.mktemp(prefix='.dea-mv-', dir=to_directory)
+        log.info("copy.put", src=from_, tmp_dest=tmp_name)
+        shutil.copy(from_, tmp_name)
+        log.debug("copy.put.done")
+        os.rename(tmp_name, to)
```
jeremyh: It's worth adding a `finally` block, similar to `copy_directory`, so that we don't leave the tmp file on disk if the copy fails mid-way-through.

jeremyh (approved):
Looks good. I particularly like the renames, they're much clearer.

I think our nc files have fletcher32 checksums (at least one of them did last time I looked), it would be worth adding that as an option in the future to support the `--checksum` option.

PR #54, opened 2017-11-20
Capture CLI `--qsub` argument

uchchwhash:
The function `capture_qsub` is supposed to return the supplied value of the `--qsub` parameter in addition to storing it in the state object. It did in its original implementation in the `agdc_statistics` but somehow after the port it is not doing that anymore.

- Return the captured `value` for the `--qsub` parameter from `capture_qsub`


jeremyh (approved):
Thanks Imam

PR #55, opened 2017-11-27
Publish DEA modules to the public directory

jeremyh:
The usage of the DEA repo has evolved, and is no longer only for private (collection management) scripts. Several other public repos now use it as a library. It should go in the same module directory as the others.


omad:
Sounds good to me. There's nothing that is necessarily private, and some tools like `stats` are now reliant on code in `digitalearthau`.

PR #59, opened 2017-12-05
Add Fractional Cover products

uchchwhash:
- Add fractional cover product definitions to the `config/products` folder
- Add product names to test


uchchwhash:
Closing this because the `fc` app can also perform the same operation.

PR #60, opened 2017-12-06
Enable batch submission

uchchwhash:
Accompanies https://github.com/GeoscienceAustralia/agdc_statistics/pull/56.

These changes are not absolutely necessary for batch
submission to work. But it plays nicer with the changes.

### DIFF
--- a/digitalearthau/qsub.py
+++ b/digitalearthau/qsub.py
@@ -35,18 +35,23 @@ class QSubLauncher(object):
     """ This class is for self-submitting as a PBS job.
     """
 
-    def __init__(self, params, internal_args=None):
+    def __init__(self, params, internal_args=None, auto_clean=None):
         """
         params -- normalised dictionary of qsub options, see `norm_qsub_params`
         internal_args -- optional extra command line arguments to add
                          when launching, particularly useful when using auto-mode
                          that passes through sys.argv arguments
+        auto_clean -- params to remove when self-submitting
         """
-        self._internal_args = internal_args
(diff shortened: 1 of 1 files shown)


digitalearthau/qsub.py
```diff
@@ -67,6 +67,7 @@ def __call__(self, *args, **kwargs):
             args = sys.argv[1:]
             args = remove_args('--qsub', args, n=1)
             args = remove_args('--queue-size', args, n=1)
+            args = remove_args('--batch', args, n=1)
```
Kirill888: This is getting uglier and uglier. When it was just `--qsub` it was okish, but now the list keeps growing and it has nothing to do with this class. This logic needs to be moved up to the caller, whoever adds internal args should also supply auto-clean list of args, or maybe caller to this function should supply it

```python
auto_clean = [ ('--batch', 1), '-v', ('--verbose',0)] 

for x in auto_clean:
    arg, n = (x,0) if isininstance(x, str) else x
    args = remove_args(arg, args, n=n)
```

Kirill888 (changes requested):

digitalearthau/qsub.py
uchchwhash: But I want the verbose flag to propagate 😕

Kirill888: I mean it as an example of input to the function, it shouldn't be defined in here, but passed in by the caller, since the set of arguments to be pruned should be caller responsibility

Kirill888 (approved):

PR #61, opened 2017-12-07
Increase memory margin per node

uchchwhash:
The memory left to the rest of the system per node (currently 512MB)
is perhaps not enough to make sure the job gets scheduled faster
according to @daleroberts.

- Change the memory request to be 1GB less per node than available
- Fix tests to reflect the change

### DIFF
--- a/digitalearthau/qsub.py
+++ b/digitalearthau/qsub.py
@@ -345,7 +345,7 @@ def norm_qsub_params(p):
     mem = normalise_mem(p.get('mem', 'small'))
 
     if nodes:
-        mem = int((mem * NUM_CPUS_PER_NODE * 1024 - 512) * nodes)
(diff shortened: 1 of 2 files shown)


digitalearthau/qsub.py
```diff
@@ -345,7 +345,7 @@ def norm_qsub_params(p):
     mem = normalise_mem(p.get('mem', 'small'))
 
     if nodes:
-        mem = int((mem * NUM_CPUS_PER_NODE * 1024 - 512) * nodes)
+        mem = int((mem * NUM_CPUS_PER_NODE * 1024 - 2048) * nodes)
```
Kirill888: How do we tune this?
Probably time to take it out into a module level constant or even into a qsub configuration option? That way we can experiment with various values without changing code each time.

PR #67, opened 2018-01-03
+ ffmpeg

benjimin:
Prerequisite for matplotlib.animate in jupyter notebooks

### DIFF
--- a/modules/py-environment/environment.yaml
+++ b/modules/py-environment/environment.yaml
@@ -17,6 +17,7 @@ dependencies:
 - distributed
 - ephem
 - fiona
+- ffmpeg
 - folium
 - future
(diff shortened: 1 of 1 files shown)


benjimin:
Here is a work-around that successfully displayed a python (matplotlib) animation in a browser:

1. `module add ffmpeg` in addition to `agdc-py3-prod`
2. `pip install --user ffmpeg-python`
3. Start a jupyter notebook and note the port number
4. Use a putty session to set up forwarding (for the current jupyter port, and current vdi host node)
5. Use local web browser (because vdi configuration of firefox does not support the same video container format which is supported by matplotlib & ipython).

Might be easier if conda also distributed web browsers?

PR #70, opened 2018-01-16
Add voluptuous to the python env on the NCI

### LINKED ISSUE #69: Add voluptuous to environment module
It's now required by stats, and the automatic deployment of stats doesn't handle installing additions dependencies.

Alternatively, fix the automatic deployment, but that's harder...

omad:
It's now required for datacube stats.

Closes #69

### DIFF
--- a/modules/py-environment/environment.yaml
+++ b/modules/py-environment/environment.yaml
@@ -79,6 +79,7 @@ dependencies:
 - sqlalchemy
 - tqdm
 - xarray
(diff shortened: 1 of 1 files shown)


jeremyh:
Close enough to abecedarian, I'll allow it 🙂

omad:
Oh shoot, I said the alphabet in my head, decided v came before x, and then put the line in the wrong place!

Thanks :)

PR #73, opened 2018-01-31
A fuse function for WOfS water layer

uchchwhash:
WOfS `water` layer is a bit-field where a bit is set to indicate `nodata`.
Hence it requires a customized fuse function (this one provided by @benjimin).


jeremyh (approved):
👍

PR #75, opened 2018-02-14
Add google analytics to DEA docs

omad:
What it says on the lid. We want to know how many people are using our manual, and where they're coming from.

Is tracked through a Geoscience Australia Google Analytics (GA GA) account.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -104,13 +104,13 @@ try:
     import sphinx_rtd_theme
     html_theme = 'sphinx_rtd_theme'
     html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+    html_theme_options = {
+        'logo_only': True,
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -104,12 +104,13 @@
...
     html_theme = 'sphinx_rtd_theme'
     html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
-    html_theme_options = {
-        'logo_only': True,
-    }
 except ImportError:
     html_theme = 'alabaster'
 
+html_theme_options = {
+    'logo_only': True,
```
jeremyh: Specifying `logo_only` breaks the doc builds for me (running `cd docs; make html`):

    preparing documents... 
    Theme error:
    unsupported theme option 'logo_only' given

I think it needs to be in the block above, so that it's only specified for the `sphinx_rtd_theme`

PR #79, opened 2018-03-06
in sync pass url to processes

uchchwhash:
Since the `Index` object should not in general be serializable (and currently it is not),
do not pass it to the forked processes in the `sync` tool. Pass the URL instead, and
recreate the `Index` inside the processes.


omad (approved):
Thanks Imam.

This looks like a pragmatic fix.

PR #81, opened 2018-03-13
[WIP] NCI module creation updates

omad:
This is an attempt to update and simplify the module deployment process at the NCI. It is powered by an entirely new script in [modules/build_environment_module.py](https://github.com/GeoscienceAustralia/digitalearthau/blob/0776996e77bea45de2da77efccda0cd2fa1b0e57/modules/build_environment_module.py)

 * Remove support for python 2 modules
 * Remove support for modules simply for different database servers
 * Move from a ~7 module deployment to only 2 modules
   1. An environment module containing all libraries and dependencies. Named `dea-env`, and versioned by date it was created `YYYYMMDD`
   2. A DEA environment containing all DEA code (digitalearthau, ODC, FC, Stats, eodatasets), named `dea` and also versioned by creation date.

The intention is to deploy a new DEA environment any time there are stable updates to anything contained within it.

### Outstanding
`dea` module needs to depend on the latest stable `dea-env` module.


modules/build_environment_module.py
```diff
@@ -0,0 +1,193 @@
...
+    return datetime.datetime.now().strftime(date_format)
+
+
+def make_output_dir(base_path) -> Path:
+    today = date()
+
+    output_dir = base_path / today
+
+    while output_dir.exists():
+        index = 1
```
jeremyh: `+= 1` ?

modules/dea/requirements.txt
```diff
@@ -0,0 +1,5 @@
+git+https://github.com/opendatacube/datacube-core.git@develop#egg=datacube
+git+https://github.com/GeoscienceAustralia/eo-datasets.git@develop#egg=eodatasets
+git+https://github.com/GeoscienceAustralia/digitalearthau.git@develop#egg=digitalearthau
```
jeremyh, line 3: I presume we'll use stable branches for all of these once finalised? (or 'nci' branches if more control is needed...)

jeremyh (approved):
huuuge improvement — nice work

modules/build_environment_module.py
omad: This function isn't used at the moment, it'll just fail if you attempt to create a two versions of a module on the same day. Simple solution is to delete the first one, or wait until tomorrow.

This was going to automatically generate a new `YYYYMMDD_{index}`, but it started feeling complicated and unnecessary.

Will probably just delete this function.

modules/dea/requirements.txt
omad: Yep, as a first step `stable` branches.

I'm also thinking about having auto-deploys for stable version releases, which might involve having Travis build a conda or python package, upload to S3, and have `requirements.txt` point there instead of git repositories. Maybe even upload to PyPI automatically and use that. Rolling towards less manual steps.

We could also easily create a configuration to do `dea/develop` style module builds, that *do* point to the develop branches of everything, and is automatically kept up to date with them.

PR #86, opened 2018-03-14
Add documentation about building and using the new modules

omad:
- Remove the duplicate release process for ODC
  We should reference it, not copy it wholesale

- Add documentation about the new modules structure
- Add a `livehtml` target to the Makefile to use while developing docs
- Add a changelog for tracking changes to the DEA environments

- Restructure and re-title some documentation sections


docs/internal/modules.rst
```diff
@@ -0,0 +1,119 @@
...
+
+    module load python3/3.6.2
+    pip3 install --user yaml
+
+Building a new *Environment Module*
+-----------------------------------
+
+::
+
+    ./build_environment_module.py
```
jeremyh: It currently needs a config file argument:
```
     ./build_environment_module.py dea-env/modulespec.yaml
```
It might be nice to have a sensible default, though?

docs/internal/release_process.rst
```diff
@@ -1,81 +0,0 @@
...
-#. Update the default version on `raijin`
-    Follow the instructions under **Update default version** in the `Datacube Environment`_ repository.
-
-#. Notify the community of the release using the Datacube Central mailing list
-    The notifications are sent out using MailChimp_. You might need an invitation from the Geoscience
-    Australia account.
-    
-    Create a campaign (possibly by replicating one of the existing ones) and change the notes.
-    You could send out a test mail to selected accounts before sending it out to the entire DEA
-    Beta Users list.
```
andrewdhicks, line 65: Mail Chimp should be moved to the DEA Release Process as well?
It is a DEA-announcement, not from ODC, specific to Raijin?

andrewdhicks (approved):
Looks great!
Only concern is the Mailchimp mailout, then we can remove steps 4 & 5 from http://datacube-core.readthedocs.io/en/latest/about/release_process.html

PR #89, opened 2018-03-20
Making `install_pip_packages` accept `prefix` or `target`

### LINKED ISSUE #88: Making modulespec.yaml-pip_cmd a little more generic
I've been implementing the `build_environment_module.py some_module/modulespec.yaml` "way of doing things" for installing my QGIS plugin on the NCI  (when conda-forge package QGIS3).

One issue I've encountered is that [`build_environment_module.py`](https://github.com/GeoscienceAustralia/digitalearthau/blob/develop/modules/build_environment_module.py#L150) has the `--prefix` arg hardcoded in the install_pip_packages function.    

This is fine for a standard `pip` install, but while my plugin is technically a python package, it is not installed to a standard location (i.e. not site-packages et al.), instead it is copied to a directory not on the python path and the modulefile sets an env var (QGIS_PLUGINPATH) to tell QGIS where to load plugins from.  A pip command that works with this would be something like `pip install -v --no-deps **--target** {target} --compile --requirement {requirements} `.  

To accommodate this sort of pip install would you consider a PR to handle both target or prefix?

I was thinking of something along the lines of:

_some_module/modulespec.yaml_

    install_pip_packages:
      requirements: requirements.txt
      target: "{module_path}"
      pip_cmd: "module load my-env; pip "

_another_module/modulespec.yaml_

    install_pip_packages:
      requirements: requirements.txt
      prefix: "{module_path}"
      pip_cmd: "module load some-env; pip "

_build_environment_module.py_

    def install_pip_packages(pip_conf, variables):
        fill_templates_from_variables(pip_conf, variables)
        pip = pip_conf['pip_cmd']
        prefix = pip_conf.get('prefix')
        target = pip_conf.get('target')
        requirements = pip_conf['requirements']
        if prefix:
            dest = prefix
            arg =  f'--prefix {prefix}'
        else:
            dest = target
            arg =  f'--target {target}'
    
        LOG.debug(f'Installing pip packages from "{requirements}" into directory "{dest}"')
        run(f'{pip} install -v --no-deps {arg} --compile --requirement {requirements}')

lpinner:
Make` install_pip_packages` a little more generic by supporting either `target` or `prefix` keys.

This allows a non-standard install location to be specified by the target key which is needed by lpinner/dea-datacube-qgis to install the datacube-qgis plugin to the $QGIS_PLUGINPATH folder via `pip install --target <path> etc...`.

Changes:

- Add  `target` key and replace `dest` key with `prefix` in `install_pip_packages`.
- Retain `dest` for backwards compatibility.
- Raise Exception if neither target or prefix or both target and prefix in config.

Notes:

 - Tested building dea and dea-env modules on Ubuntu 17.10 with environment-modules 3.2.10. Not tested on NCI.
 - I couldn't find any tests or docs for the modules (other than how to run them), so haven't added/updated that side of things.
- No checks are made to ensure `target` path exists, pip does that for us - https://github.com/pypa/pip/blob/master/src/pip/_internal/commands/install.py#L363

Closes GeoscienceAustralia/digitalearthau#88

### DIFF
--- a/modules/build_environment_module.py
+++ b/modules/build_environment_module.py
@@ -143,11 +143,20 @@ def fix_module_permissions(module_path):
 def install_pip_packages(pip_conf, variables):
     fill_templates_from_variables(pip_conf, variables)
     pip = pip_conf['pip_cmd']
-    dest = pip_conf['dest']
+    prefix = pip_conf.get('prefix',pip_conf.get('dest'))  # 'dest' for backwards compatibility
+    target = pip_conf.get('target')
     requirements = pip_conf['requirements']
-    LOG.debug('Installing pip packages from "%s" into directory "%s"',
-              requirements, dest)
-    run(f'{pip} install -v --no-deps --prefix {dest} --compile --requirement {requirements}')
+    if prefix and target is None:
+        dest = prefix
+        arg = f'--prefix {prefix}'
+    elif target and prefix is None:
+        dest = target
+        arg = f'--target {target}'
+    else:  # Either no target or prefix OR target and prefix were in the conf
+        raise Exception('Either prefix: <prefix path> or target: <target path> is required by install_pip_packages:')
+
+    LOG.debug(f'Installing pip packages from "{requirements}" into directory "{dest}"')
+    run(f'{pip} install -v --no-deps {arg} --compile --requirement {requirements}')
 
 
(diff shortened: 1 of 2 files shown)


omad (approved):
Nice. Thanks Luke.

PR #94, opened 2018-04-05
NCI module improvements

omad:
- Fix an error on module load about PYTHONPATH already being set, even when it wasn't
- Add a script for automatically running a _requirements met_ jupyter notebook as a PBS job
- Fix the installation of jupyter lab extensions into the `dea-env` module
  - And a few interesting looking extensions
- Make the `dea` module depend on a specific version of `dea-env`, not simply the latest version
- Some incomplete work on a script to automatically test a few things about a newly created module


jeremyh (approved):
:+1:

PR #97, opened 2018-04-11
Add legacy PQ products for WOfS

andrewdhicks:
Damien says:
> Legacy PQ is used in the generation of WOfS. We need to index it to have accurate provenance tracking for WOfS. The scene data will be deleted once WOfS has been generated.


bellemae (approved):
Looks good, thanks!

omad (approved):

PR #98, opened 2018-04-12
More env fixes

omad:
- Fix the pycodestyle breakage
- Upload test coverage results to codeclimate from Travis CI
- Use s3pypi to upload artefacts to S3 in a PyPI style fashion

Can use with `pip install digitalearthau --extra-index-url https://d2mru5c1aqk868.cloudfront.net/`


setup.py
```diff
@@ -18,6 +18,7 @@
 
     url='https://github.com/GeoscienceAustralia/digitalearthau',
     author='Geoscience Australia',
+    author_email='damien.ayers@ga.gov.au',
```
andrewdhicks, line 21: You could also use DEADevelopers@GeoscienceAU.onmicrosoft.com

modules/dea_stable/modulespec.yaml
```diff
@@ -0,0 +1,32 @@
+variables:
+  module_name: dea
+  module_description: DEA tools for the NCI
+  modules_base: "/g/data/v10/public/modules"
+  dbhost: 130.56.244.105
```
andrewdhicks, line 5: Better to use hostname, in case the IP changes?

andrewdhicks (approved):

setup.py
omad: That is a truly awful email address! It starts with *DEAD*, wtf is `onmicrosoft.com`, *shivers*.

Maybe

modules/dea_stable/modulespec.yaml
omad: Two things to fix first:
1) Get the NCI to rename hosts from `agdc-*` to `dea-*`
2) Write a script to update everyone's `.pgpass` to use the hostname, and maybe add support for `staging` and `dev`.

setup.py
omad: I added an email address because `s3pypi` complained if there wasn't one. It isn't getting sent to PyPI or anything.

PR #99, opened 2018-04-12
Document other modules on raijin

andrewdhicks:
Document pycharm and nctools


docs/internal/other_modules.rst
```diff
@@ -0,0 +1,44 @@
...
+Pycharm (Community Edition) is a free
+The easiest way to run Pycharm is to copy the launcher on to your desktop.
+
+You can do this by running:
+
+    cp /g/data/v10/public/modules/pycharm/PyCharm.desktop ~/Desktop/
+
+
+To upgrade the version, download the latest tarball from the jetbrains website and place it in the folder
+``/g/data/v10/public/modules/pycharm/``
```
omad, line 21: It'd be good to mark this as **for administrators only**, or move it somewhere else.

I think splitting documentation by intended user and task will make it more useful, rather than by software module.

omad (approved):
Awesome!

I'm going to merge because this is an awesome improvement, while leaving a note about how we can improve it further.

ashoka1234 (commented):
Looks good. Agree with Damien on administrator specific aspects.

docs/internal/other_modules.rst
andrewdhicks: Since it is only one line, I'll make it as admin only. If it gets any bigger, I'll move it to the internal section.

PR #101, opened 2018-04-17
Awalsh source products type consistency

walsh-andrew:
In DEA stacker.py there is an inconsistency with ODC:
output_product vs output_type (respectively).

Change DEA stacker.py to refer to output_type


omad (approved):
So simple, so critical. Thanks Andrew

PR #104, opened 2018-04-26
Fix dea-submit-sync

omad:
It was hard coded to expect an environment to be already set up inside
PBS jobs. Presumably expecting that `module load` commands were run inside
shell init scripts?

Either way, hopefully this isn't too hard coded, but it has the benefit of being simple.

### DIFF
--- a/digitalearthau/sync/submit_job.py
+++ b/digitalearthau/sync/submit_job.py
@@ -18,7 +18,7 @@ from boltons import fileutils
 from click import style
 
 from datacube.index import index_connect
-from digitalearthau import collections
+from digitalearthau import collections, MODULE_NAME, MODULES_PATH
 from digitalearthau.collections import Trust
 from digitalearthau.paths import get_dataset_paths
 from digitalearthau.sync import scan
@@ -163,6 +163,8 @@ class SyncSubmission(object):
             '-o', str(output_file),
             '-W', ','.join(attributes),
             '--',
+            'module use %s; ' % MODULES_PATH,
+            'module load %s; ' % MODULE_NAME,
             *sync_command
         ]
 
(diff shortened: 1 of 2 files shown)


jeremyh:
It's currently using `qsub -V`, which passes all your existing (on submission) environment variables through to the job (ie: your loaded modules/pythonpaths). That was a simple way of supporting submission for all module and non-module-PYTHONPATH usage, as you already have the right module loaded (or dev setup) to run the `dea-submit-sync` tool itself.

I've always been slightly hesitant about using `-V` though, as users themselves could have strangely-imported modules or env variables, causing running jobs to have minor differences between each other. Explicit module imports may be more consistent, though there's admittedly still a risk through different user bash profiles.

I'm happy to change it, but we should choose one strategy: explicit modules, or `-V` :)

digitalearthau/sync/submit_job.py
```diff
@@ -163,6 +163,8 @@ def submit(self,
             '-o', str(output_file),
             '-W', ','.join(attributes),
             '--',
+            'module use %s; ' % MODULES_PATH,
+            'module load %s; ' % MODULE_NAME,
```
jeremyh, line 167: I haven't run it, but this may not work? I think qsub takes a command, not a bash string.

Edit: from the man page: `-- executable  [arglist for executable]]`

jeremyh: From the man page: `-- executable  [arglist for executable]]`

or else specify a script path without `--`

omad:
This was a disaster of a Pull Request, sorry.

The problem turned out to be that it's not possible to run `dea-submit-sync` from VDI, since the current environment variables aren't submitted with the `qsub -V` option. For now this can be solved with documentation and communication, so I'll close this PR.

Thanks.

PR #106, opened 2018-05-02
Asv/s2 nrt product definitions

ASVincent:
* Adds metadatatype sentinel_2_alpha which is the eo metadatatype + fallback to center_dt
  * This definition can be expanded/changed in the future, it's just a base line for the alpha products of sentinel 2
* Adds product types for:
  * level 1 scenes from the s2_aws_pds
  * level 2 scenes for nrt; currently these scenes are identified by lineage, this should also change in the future.


digitalearthau/config/products/s2_nrt_scenes.yaml
```diff
@@ -0,0 +1,1726 @@
...
+metadata:
+    format:
+        name: JPEG2000
+    instrument:
+        name: MSI
+    platform:
+        code: SENTINEL_2A
+    processing_level: Level-1C
+    product_type: level1
+    product_format:
```
harshurampur, line 18: Do we need product_format and lineage in the product definitions below added ?

harshurampur (approved):

ASVincent:
Product definitions will go to our ecs mgmt repository for now.

PR #107, opened 2018-05-02
Fix Ls8-PQ-legacy collections

ashoka1234:
Copy paste error in Ls8-PQ-legacy collection in collections.py


andrewdhicks (approved):
My bad, sorry!

PR #108, opened 2018-05-07
Deploy to s3 using build in travis tools

omad:
The s3pypi package had problems with our versioning scheme and
has proved itself unsuitable.


andrewdhicks (approved):
That's a shame that s3pypi wasn't suitable.
Thanks for trying to get it working!

PR #109, opened 2018-05-08
Fix Package Deployments from Travis-CI

omad:
Travis-CI by default only clones 50 levels deep of a repository, which means
our fancy version numbering support wasn't working, creating package names 
like digitalearthau-0+untagged.50.g59ebe36.dirty.tar.gz .

This PR makes Travis clone 9999 levels deep so that we always get a version 
number.

It also starts uploading our test coverage results to codecov.io as well
as to coveralls, since I think it's a better service, and we can run
them both in parallel to compare.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ language: python
 python:
     - '3.6'
 sudo: true
+git:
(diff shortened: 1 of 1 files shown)


andrewdhicks (approved):
Looks great!

PR #110, opened 2018-05-08
Change the way docs and package building happens on TravisCI

omad:
The TravisCI build has been responsible for building and deploying the
DEA technical documentation via a `docs/deploy.sh` script. Unfortunately
this build process leaves the git tree in a _dirty_ state, which causes
the generated package names to include the word _dirty_ in them. Not exactly
what we want.

This PR switches to building the documentation in the `build` phase of the
TravisCI run, and deploys it to `gh-pages` using the [GitHub Pages Deployment](https://docs.travis-ci.com/user/deployment/pages/)
available from Travis. It seems configurable and able to meet our needs,
so we will hopefully be able to remove the `docs/deploy.sh` if it proves
successful.

This change fixes the naming of the python packages.


.travis.yml
```diff
@@ -89,6 +91,16 @@ deploy:
...
 
+ -  provider: pages
+    skip-cleanup: true
+    github-token: $GITHUB_TOKEN  # Set in the settings page of your repository, as a secure variable
+    local-dir: docs/_build/html/
+    keep-history: true
+    committer-from-gh: true
+    verbose: true
+    on:
+      branch: develop
```
andrewdhicks, line 102: Add `repo: GeoscienceAustralia/digitalearthau` to the conditional trigger, or would we not expect/support forks?

andrewdhicks (approved):
This is much better than the script! 
At the time, it was all about the git commits needing to protect the Personal Access Token. (From https://gist.github.com/domenic/ec8b0fc8ab45f39403dd#get-encrypted-credentials) 
As long as we are happy it isn't an issue here, being a GA-only repo, it looks great imrpovement!

.travis.yml
omad: Fixed, I always forget that bit.

omad:
I don't think using this Travis feature is any less safe than using the script. Both require access to a Personal Access Token.

andrewdhicks:
The script doesn't use a Personal Access Token, it uses a SSH deploy key with permissions limited to that repo only, rather than any repo by the owner of the Token.

omad:
Aaah, okay, I'll grimace a bit, but it should be okay.

PR #111, opened 2018-05-08
No more submodules

omad:
`git submodules` weren't really the right choice for including the `dea-notebooks` repo when building the documentation. They point to a specific commit, whereas we always want to pull in `HEAD`. They just add complexity without us benefitting from them.

This PR replaces the `git submodule` for `dea-notebooks` with an additional target in `docs/Makefile` called `fetchnotebooks` which clones the `dea-notebooks` repo into the correct location. It can be called when building the documentation like `make fetchnotebooks html`.

I've also added `docs/notebooks/` to the `.gitignore` so that it doesn't accidentally get checked in, and updated the build on Travis-CI to use the new `Makefile` target.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,8 @@ docs/html/
 
 # Generated Documentation
 docs/dev/generate/
+# Notebooks are pulled from the dea-notebooks repo
+docs/notebooks
 
 # MYPY static type checker cache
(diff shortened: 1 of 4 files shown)


andrewdhicks (approved):
Looks great! Much cleaner!

PR #114, opened 2018-05-14
PQ stats collections missing index connection

uchchwhash:
Not completely understanding the collections API, I forgot to set the index for the PQ summary collections, and `dea-sync` understandably failed. This PR fixes this issue.


jeremyh (approved):
Thanks Imam. 

(this function is getting more unwieldy and error prone than I'd like :/ The new FC collection definitions don't receive the index either)

PR #122, opened 2018-06-13
Make DEA test fixtures configurable

jeremyh:
Our testing plugin currently provides a few pytest fixtures for testing with a real datacube index:
```python
pytest_plugins = "digitalearthau.testing.plugin"

def test_dc_with_dea_config(dea_index):
    assert dea_index.products.get_by_name("ls7_nbar_albers") is not None
```

These are automatically initialised and torn down for every individual test.

The dashboard uses these, but its tests have expensive setup: it loads a lot of test data, which takes a while, and it's not modifed within the tests themselves. It would be nice to have a single session-scoped instance of datacube that runs this data-load once, rather than repeatedly for every test.

These commits add factory methods for creating custom pytest fixtures, where you can specify your own scope or datacube config to use.

As an example, the dashboard can now define session-scoped fixtures like this in its `conftest.py`:
```python
from digitalearthau.testing import factories

pytest_plugins = "digitalearthau.testing.plugin"

# Use session-scoped databases, as it takes a while to populate with
# our data, and we're treating it as read-only in tests.
session_db = factories.db_fixture('local_config', scope='session')
# (notice it's using the above session_db)
session_index = factories.index_fixture('session_db', scope='session')
session_dea_index = factories.dea_index_fixture('session_index', scope='session')
```

---
Another minor fix included here: the default integration test config wasn't included in the manifest, so an error was thrown if you were using an installed instance of DEA.


omad:
Hey Jeremy, I totally understand the need for getting tests running faster, but I don't think this is the best way to achieve it.

The test infrastructure used with `datacube`, `digitalearthau` and `cubedash` has plenty of room for improvement. I think it should be along the lines of consolidating useful functions in `datacube`, and simplifying things elsewhere. One of the aims should be to make it much easier to write integration tests. It would be great to be able to rollback database changes to make tests run faster, but I'm not sure if this is feasible.

My main objections would be:
 - Adding factory methods to generate pytest fixtures complicates an already somewhat difficult to work with feature of the tests.
 - `cubedash` only depends on `digitalearthau` for setting up a test database. But I think it should depend only on `datacube`, and be self contained with any extra test data.

jeremyh:
Agreed that we should consolidate the testing infrastructure to datacube. I put it in the DEA repo so that we could stabilise the API (and figure out how to actually design it through usage) before adding it to the core (where we'll feel more pressured to maintain API stability). I'll happily move `cubedash` as soon as the equivalent functionality is in core -- as you said, the dependency on DEA is only within its tests.

A transaction/rollback model would be good too: but there's no simple way that I found to do it, as the various datacube functions and tools all begin/commit their own transactions, and the cli tools have their own connections too, all of which we mix and match within tests. The best way I can think of to do it would require complicating the core db drivers, and contain various asterisks in its usage directions (no forking cli commands etc)

The factory model used here matches most library pytest plugins I've found, see pytest_postgresql as an example: https://pypi.org/project/pytest-postgresql, pytest doesn't seem to offer a clean alternative from my searching: the user is literally defining new fixtures, not configuring existing ones.

These changes, at least, are opt-in in the sense that they maintain the simplicity of the original pytest plugin: with one line of code you get a datacube to use in integration tests, safely with no extra configuration. You don’t need to know about the factories unless you’re defining new fixtures.

The single biggest reason for adding these as a pytest plugin was avoiding the copying-pasting of all of this integration test code, and user configuration, across many repositories. It's delicate and likely to change as the ODC internal implementations change. And currently any deviation at all (such as needing a second instance in a test, or a different scope) needs us to fall back to copy-pasting the whole lot.

omad (approved):
Okay, using `digitalearthau` as a proving ground for test code makes me happy.

I know all about the issues with database rollbacks. Maybe at some point we should do some rearchitecting of the database connection handling used throughout to make it all more testable.

PR #123, opened 2018-06-13
Fix FC sync check issue and create work directory when it does not exist

santoshamohan:
Updated the following files:
**1) digitalearthau/collections.py:** To fix FC integrity check issue.
**2) digitalearthau/paths.py:** To notify user and create one, when NCI root work directory does not exists. Also, rename base path only when trash path exists.
**3) digitalearthau/sync/submit_job.py:** Create run path when it does not exists. Notify user if the run path exists.

### DIFF
--- a/digitalearthau/collections.py
+++ b/digitalearthau/collections.py
@@ -380,15 +380,14 @@ def init_nci_collections(index: Index):
             )
         )
 
-    def add_fc_albers_collections(name: str, project='fk4'):
+    def add_fc_albers_collections(project='fk4'):
         _add(
             Collection(
-                name='ls5_{}_albers'.format(name),
-                query={'product': 'ls5_{}_albers'.format(name)},
+                name='ls5_fc_albers',
+                query={'product': 'ls5_fc_albers'},
                 file_patterns=(
                     '/g/data/{project}/datacube/002/FC/'
-                    'LS5_TM_{name}/*_*/LS5*{name}*.nc'.format(project=project,
-                                                              name=name.upper()),
+                    'LS5_TM_FC/*_*/LS5*FC*.nc'.format(project=project),
                 ),
                 unique=('time.lower.day', 'lat', 'lon'),
                 index_=index,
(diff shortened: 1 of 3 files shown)


digitalearthau/paths.py
```diff
@@ -9,10 +9,12 @@
 
 import pathlib
 import structlog
+import logging
 
 from datacube.utils import is_supported_document_type, read_documents, InvalidDocException, uri_to_local_path
 
 _LOG = structlog.getLogger()
+_LOG1 = logging.getLogger(__file__)
```
omad: I don't think it makes sense to log in both plain text and structured forms. It'd be better to stay consistent with what is here and use the `structlog`.

digitalearthau/sync/submit_job.py
```diff
@@ -354,10 +354,10 @@ def _find_and_submit(tasks: List[Task],
...
 
         run_path = task.resolve_path(work_folder).joinpath('{:03d}'.format(submitted))
-        if run_path.exists():
-            raise RuntimeError("Calculated job folder should be unique? Got %r" % (run_path,))
-
-        fileutils.mkdir_p(run_path)
+        if not run_path.exists():
+            fileutils.mkdir_p(run_path)
+        else:
+            _LOG.warning("Calculated job folder should be unique? Got %r", run_path)
```
omad, line 360: Was this causing errors with testing? What's the reasoning for changing from an Error to a logged warning?

digitalearthau/collections.py
```diff
@@ -384,47 +380,63 @@ def add_albers_collections(name: str, project='rs0'):
...
+    # PQ/NBAR/NBART Albers
+    add_albers_collections('pq')
+    add_albers_collections('nbar')
+    add_albers_collections('nbart')
+
+    # FC
+    # Example: ingested fractional cover:
+    # LS5_TM_FC  LS7_ETM_FC  LS8_OLI_FC
+    # /g/data/fk4/datacube/002/FC/LS5_TM_FC/13_-22/LS5_TM_FC_3577_13_-22_20030901235428500000_v1490733226.nc
+    add_fc_albers_collections('fc')
```
omad: If this function is only for `FC`, it would be simpler and nicer to not include the `name` argument.

omad (changes requested):
Thanks, just a couple of things to fix before we merge.

digitalearthau/paths.py
santoshamohan: Ok. I will modify the code to be consistent with structured form.

digitalearthau/sync/submit_job.py
santoshamohan: Upon multiple execution of same script, earlier code was raising an execution when the run_path already existed. Test were aborted due to this exception. To avoid this, updated code shall notify user that the run_path needs to be unique.

digitalearthau/collections.py
santoshamohan: Updated the code by removing name argument.

omad (approved):

PR #124, opened 2018-06-14
Add nci env dea-module test scripts

santoshamohan:
Added new shell scripts and dependency files for testing dea module.


scripts/nci_env_deatest/datacube_config.conf
```diff
@@ -0,0 +1,12 @@
+[datacube]
+db_hostname: agdcdev-db.nci.org.au
+db_port: 6432
+db_database: ncitest_datacube_db
+db_username: sm9911
```
omad: We probably don't need to specify the username here, to make it more easily run-able by others.

scripts/nci_env_deatest/Test_DEA_Module.sh
```diff
@@ -0,0 +1,188 @@
...
+sh "$HOMEDIR"/dea_testscripts/index_and_ingest.sh "$MODULE" "$CONFIGFILE" "$HOMEDIR" > "$HOMEDIR"/output_files/idx_ingest/Index_Ingest_NCI_PC.log
+
+echo
+echo "==================================================================="
+echo "|  Test the following on Raijin system using PBS:                 |"
+echo "|  1) Indexing  LS7_ETM_NBAR/NBART, LS8_OLITIRS_NBAR/NBART Scenes |"
+echo "|  2) And then ingesting the products to test database            |"
+echo "==================================================================="
+echo
+qsub -v MUT="$MODULE",CONFIG_FILE="$CONFIGFILE",FL_PATH="$HOMEDIR",DC_CONF="$DCCONF" "$HOMEDIR"/dea_testscripts/pbs_script.sh
```
omad: What does `FL_PATH` stand for?

omad (changes requested):
This looks really really good, thank you Santosh!

Just a few things before we merge it:
- It would be great to run the tests as different users. It looks to be almost possible, but still has your username/paths hard coded in a couple of places.
- I think it would be better to not duplicate the FC/Ingestion/Stats configuration files. Instead, i It should be possible to copy them from the module under test, and run a `sed -e -e 's/path/newpath/g` replacement over them. I'd expect this to be lower maintenance in the future and a slightly more thorough test in case the configurations are ever broken.
- Can you please include a README with instructions for how to run this. I'm not sure whether the intention is to copy this directory somewhere and run it, or... something else.

Cheers! 👍

scripts/nci_env_deatest/datacube_config.conf
santoshamohan: Thanks for your feedback, Damien. 
I will make changes to the scripts as mentioned, to make it more generic and useable.
Also, I will add a README.md file and remove the username from the config file.

scripts/nci_env_deatest/Test_DEA_Module.sh
santoshamohan: FL_PATH means File Path. Do you want me to change this to FILE_PATH?

omad: I'm not sure `FILE_PATH` is meaningful either. Maybe something like `TEST_BASE` or `TEST_PATH`? Depends on how and what it's used for.

santoshamohan: TEST_BASE would be more appropriate. I will update FL_PATH to TEST_BASE.

santoshamohan:
Hi Damien,

I have update the comments and pushed updated code to the repo.

omad (approved):
Thanks

omad (approved):

PR #125, opened 2018-06-14
fix naming of nir bands

felipk101:
Product definition for sentinel mismatch with s3 metadata store for NIR bands causes NIR1 and NIR2 not to load.

### DIFF
--- a/digitalearthau/config/products/ga_s2_ard.yaml
+++ b/digitalearthau/config/products/ga_s2_ard.yaml
@@ -525,7 +525,7 @@ measurements:
     fmask:
       bits: [0, 1, 2, 3, 4, 5, 6, 7]
       description: Fmask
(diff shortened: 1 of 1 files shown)


omad:
I don't know much about this, but can you please have a look @roarmstrong.

harshurampur (commented):
Thanks Felix, the changes needs to be done to nbart product as well .. can you please make those changes for the band name starting as `t_nir` and `t_rededge4`..?

felipk101:
Sure will do...

harshurampur (changes requested):
There are some other changes with respect to the pixel quality(CFmask) that has to be done before merging, let me know if I can do those changes here.

harshurampur (approved):
Awesome! Thanks

PR #131, opened 2018-07-18
Remove format in geomedian product definition

harshurampur:
Remove `format` in metadata column to index netcdfs for geomedian annual product


uchchwhash (approved):
I approve! But can you merge it since the build is failing?

PR #132, opened 2018-07-18
Fix Travis-CI breakage

omad:
The Travis-CI build is broken, which prevents us merging anything.

The two breakages which this PR addresses are:
- Roll back to an old version of `datacube-core` with the old indexing behaviour, which is used in some tests. `datacube-core` is being worked on today by @Kirill888 to turn the new indexing behaviour into an API which `digitalearthau` can switch to as soon as it is available.
- PyLint 2.0.0 was released, which includes new checks. I've fixed a couple of things, and changed our pylint configuration to be a white-list of checks, so that new releases of pylint won't default to breaking our builds. (hopefully)


Kirill888 (approved):
👍

PR #137, opened 2018-07-25
Add dumps of FC and WOfS products

omad:
WOfS and FC apps generate their own product definitions into the production database.

Include a dump of them in the github repo for reference purposes, and for anyone indexing our data into a new database.


andrewdhicks (approved):
+1

PR #139, opened 2018-07-27
Update dea-coherence for removing duplicate datasets in 1:1 relationships

santoshamohan:
When creating derived products in DEA, some have a simple 1:1 relationship with their parent product. 
For example: every NBART Albers Tile, should have 1 and only 1 Fractional Cover tile.

Our current processes cover the case where a tile isn’t generated, and will fill in the gaps.
Sometimes, processing will create more than one tile, and we need to be able to clean these up in a safe and automated manner. At the moment, we have duplicate LS7 FC tiles around the end of 2017 (See https://data.dea.gadevs.ga/ls7_fc_albers/2017 the graph from the data dashboard)

The dea-coherence tool is able to find these 'sibling' datasets, but has no way of dealing with them.

**Proposed changes**
Modify dea-coherence tool to be able to choose the most recent sibling and archive the others.

The command to list FC duplicates:
- `dea-coherence --check-ancestors --check-siblings product=ls7_fc_albers '2017-10-01 < time < 2017-10-15'`
    
The command to archive FC duplicates:
- `dea-coherence --archive-siblings product=ls7_fc_albers '2017-10-01 < time < 2017-10-15'`

### DIFF
--- a/digitalearthau/coherence.py
+++ b/digitalearthau/coherence.py
@@ -67,19 +67,19 @@ def main(expressions, check_locationless, archive_locationless, check_ancestors,
             if check_locationless or archive_locationless:
                 if dataset.uris is not None and len(dataset.uris) == 0:
                     locationless_count += 1
-                    
+
                     if archive_locationless:
                         dc.index.datasets.archive([dataset.id])
                         archive_count += 1
                         _LOG.info("locationless_dataset_id.archived", dataset_id=str(dataset.id))
                     else:
                         _LOG.info("locationless_dataset_id", dataset_id=str(dataset.id))
-            
+
             # If an ancestor is archived, it may have been replaced. This one may need
             # to be reprocessed too.
             if check_ancestors or archive_siblings or check_siblings:
                 archive_count += _check_ancestors(check_ancestors, check_siblings, archive_siblings, dc, dataset)
-        
+
         _LOG.info("coherence.finish",
                   datasets_count=count,
                   locationless_count=locationless_count,
@@ -90,15 +90,15 @@ def main(expressions, check_locationless, archive_locationless, check_ancestors,
 def _archive_dataset_ids(dc, ds_id, sibling_ids):
     # sibling/s indexed_time dictionary
     sb_itime_dict = dict()
-    
+
     # Count of dataset id's archived
     archive_count = 0
-    
+
     # dataset indexed time
     ds_itime = dc.index.datasets.get(ds_id).indexed_time
     for key, s_id in enumerate(sibling_ids):
         sb_itime_dict[s_id] = dc.index.datasets.get(s_id).indexed_time
-    
+
     # new dataset id to retain
     new_id = ds_id
     for id, itime in sb_itime_dict.items():
@@ -106,18 +106,18 @@ def _archive_dataset_ids(dc, ds_id, sibling_ids):
         if ds_itime <= itime:
             ds_id_archive = new_id
             ds_archive_time = ds_itime
-            
+
             # Latest sibling dataset id to retain
             new_id = id
             ds_itime = itime
         else:
             ds_id_archive = id
             ds_archive_time = itime
-        
+
         dc.index.datasets.archive([ds_id_archive])
         _LOG.info("\tarchived_dataset_id", id=str(ds_id_archive), indexed_time=str(ds_archive_time))
         archive_count += 1
-    
+
     _LOG.info("\trecent_dataset_id", id=str(new_id), indexed_time=str(ds_itime))
     return archive_count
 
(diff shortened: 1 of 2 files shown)


digitalearthau/coherence.py
```diff
@@ -40,63 +47,121 @@ def main(expressions, archive_locationless, debug, check_locationless, check_anc
...
+
+        _LOG.info("coherence.finish",
+                  datasets_count=count,
+                  locationless_count=locationless_count,
+                  siblings_count=_siblings_count,
+                  archived_count=archive_count)
+
 
-        _LOG.info("coherence.finish", count=count, archive_count=archive_count)
+def _archive_dataset_ids(dc, ds_id, sibling_ids):
```
omad: This function should have a clearer name and some documentation. It can also be made simpler by combining `ds_id` and `sibling_ids` into a single argument, since from the point of view here, there's no useful reason to split them.

I got carried away a little and made a simpler implementation here:
```python
def _archive_duplicate_siblings(dc, ids):
    """Archive old versions of duplicate datasets.
    
    When given a list of duplicate sibling datasets, keep the most recently
    indexed one and delete all the older ones.
    
    Return the number of archived duplicates.
    """
    # Look up the indexed time for each datset and store in a dict
    id_to_index_time = {ds_id: dc.index.datasets.get(ds_id).indexed_time
                        for ds_id in ids}
    
    # Sort by indexed time, and split into [newest : older_duplicates]
    newest_ds, *older_duplicates = OrderedDict(sorted(id_to_index_time.items(), key=lambda t: t[1], reverse=True))
    
    dc.index.datasets.archive(older_duplicates)

    _LOG.info("dataset_id.archived", ids=older_duplicates)
    _LOG.info("dataset_id.kept", id=newest_ds)
    
    return len(older_duplicates)
```

digitalearthau/coherence.py
```diff
@@ -40,63 +47,121 @@ def main(expressions, archive_locationless, debug, check_locationless, check_anc
...
+                     archive_siblings: bool,
+                     dc: Datacube,
+                     dataset: Dataset):
+    global _siblings_count
+    ancestors_archive_count = 0
     dataset = dc.index.datasets.get(dataset.id, include_sources=True)
     if dataset.sources:
         for classifier, source_dataset in dataset.sources.items():
-            if source_dataset.is_archived:
+            if check_ancestors and source_dataset.is_archived:
```
omad: Would it make sense to not pass `check_ancestors` in here, and do this check regardless?

If the ancestor is _archived_ we probably don't care about duplicate siblings and would want to just archive them all.

digitalearthau/coherence.py
```diff
@@ -40,63 +47,121 @@ def main(expressions, archive_locationless, debug, check_locationless, check_anc
...
+    # sibling/s indexed_time dictionary
+    sb_itime_dict = dict()
 
+    # Count of dataset id's archived
+    archive_count = 0
 
-def _check_ancestors(check_siblings: bool, dc: Datacube, dataset: Dataset):
+    # dataset indexed time
+    ds_itime = dc.index.datasets.get(ds_id).indexed_time
+    for key, s_id in enumerate(sibling_ids):
```
omad: It's possible to iterate directly over `sibling_ids`, the `key` and `enumerate()` aren't needed here.

digitalearthau/coherence.py
```diff
@@ -40,63 +47,121 @@ def main(expressions, archive_locationless, debug, check_locationless, check_anc
...
+
+            # Latest sibling dataset id to retain
+            new_id = id
+            ds_itime = itime
+        else:
+            ds_id_archive = id
+            ds_archive_time = itime
+
+        dc.index.datasets.archive([ds_id_archive])
+        _LOG.info("\tarchived_dataset_id", id=str(ds_id_archive), indexed_time=str(ds_archive_time))
```
omad: The `_LOG` in this module is a structured logger, which is intended to be able to output logs in an easily machine readable format. It would be better not to use `\t` or similar characters in the *key* of the structured log message.

omad (dismissed):
Thanks for this new feature, and for cleaning things up a bit while you were here.

At the least I'd like a little more internal documentation before merging. I also got a bit carried away and included a (in my mind) simpler implementation of part of it to include if you want.

digitalearthau/coherence.py
santoshamohan: Replaced previous function with _archive_duplicate_siblings.

digitalearthau/coherence.py
santoshamohan: Removed check_ancestors check in the if condition

digitalearthau/coherence.py
santoshamohan: This condition is replaced by new function.

digitalearthau/coherence.py
santoshamohan: This code is replaced by new function _archive_duplicate_siblings.

santoshamohan:
Updated as per review comments

omad (approved):
Thanks!

PR #141, opened 2018-07-30
Update changes due to refactor dataset add data flow

santoshamohan:
Reason for this pull request:
During sync process of datasets using dea-sync, attribute error **AttributeError: module 'datacube.scripts.dataset' has no attribute 'load_datasets'** was raised.
This error was due to changes for refactoring of dataset add data workflow in **datacube.scripts.dataset.py** file.

Proposed changes

1. Modify **add_dataset** function to be inline with changes to **datacube.scripts.dataset.py** file.
2. Improve logging in **_find_uri_mismatches** function within /**digitalearthau/sync/scan.py** file.

### DIFF
--- a/digitalearthau/index.py
+++ b/digitalearthau/index.py
@@ -3,14 +3,12 @@ import structlog
 
 from datetime import datetime
 from typing import Iterable
-from datacube.ui.common import ui_path_doc_stream
 from datacube.index import Index
 from datacube.model import Dataset
-from datacube.scripts import dataset as dataset_script
 from datacube.utils import uri_to_local_path
 from digitalearthau.utils import simple_object_repr
-from datacube.index.hl import Doc2Dataset
-
+from datacube.ui.common import ui_path_doc_stream
+from datacube.index.hl import Doc2Dataset, check_dataset_consistent
 
 _LOG = structlog.getLogger('dea-dataset')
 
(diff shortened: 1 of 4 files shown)


omad:
Can you please have a look at this @Kirill888 . I think it should be possible to make this update without importing anything from `datacube.scripts` now, but don't know the exact details.

Thanks

digitalearthau/index.py
```diff
@@ -65,21 +70,23 @@ def add_dataset(index: Index, dataset_id: uuid.UUID, uri: str):
...
-    else:
-        raise RuntimeError('Dataset not found at path: %s, %s' % (dataset_id, uri))
+    yaml_path = uri_to_local_path(uri)
+
+    ds_resolve = Doc2Dataset(index)
+    doc_stream = ui_path_doc_stream([yaml_path], logger=_LOG, uri=True)
+    dss = dataset_script.dataset_stream(doc_stream, ds_resolve)
+
+    try:
+        dataset_script.index_datasets(dss,
```
Kirill888: `index_datasets` isn't doing much, it's really just `map(index.datasets.add, dss)` with some exception guards and logging. I would prefer we didn't depend on functionality inside scripts, they are not meant to have stable APIs. This is the case where code duplication is preferable to extra dependencies.

Kirill888 (dismissed):
We should try to get rid of all imports from `datacube.scripts`

Also is this the only place where such imports happen?

Probably worthwhile changing them all in one pull request.

Kirill888:
Actually only other import from scripts is ingest, we can leave that one as is for now.

santoshamohan:
All references to datacube.scripts import has been removed.

omad (approved):

digitalearthau/index.py
santoshamohan: Updated the code to remove index_datasets

PR #144, opened 2018-07-31
Update test datacube configuration path

santoshamohan:
Test datacube configuration path used in some of the scripts were hard-coded. Update is required to make it more generic so that scripts can be executed from any directory.

**Proposed solutions**
1. Remove dc configuration paths that are hard-coded.


omad (approved):
I'm not entirely sure about the hard coded references to a particular subdir of $HOMEDIR either, but whatever works for now is good.

PR #147, opened 2018-08-01
Set click path exists option in dea-submit-ingest qsub as false

santoshamohan:
Reason for this pull request
Setting dea-submit-ingest qsub path.exists as true would make user to enter the config options as a mandatory argument. 

Proposed solution
Set dea-submit-ingest qsub path.exists option to false


omad (approved):
Thanks!

PR #148, opened 2018-08-03
Add ingest qsub queue size command option

santoshamohan:
dea-submit-ingest command was limited to processing only 3200 tasks. 

**Proposed solution**
Add queue size command option to aid flexibility of choosing qsub task queue size.


omad (approved):
Thanks

PR #149, opened 2018-08-03
Add some extra logging to `run_distributed.sh`

omad:
When running qsub jobs we weren't logging enough information to debug some
issues.

This is a small step in the right direction...


santoshamohan (approved):
Thanks, Damien!

PR #150, opened 2018-08-03
Re-skin and restructure the DEA usage guide

omad:
View a preview of the changes at https://damien-dea-docs.netlify.com/

- Re-skin with a DEA appropriate colour scheme
- Use a scaleable logo which looks much better on high res screens
- Make all the awesome notebook/user guides more discoverable


docs/index.rst
```diff
@@ -39,20 +39,22 @@ Publicly available data access and web services are currently in development.
...
+   notebooks/Integrating_external_data/README.rst
+   notebooks/Index_calculation/README.rst
+   notebooks/Temporal_analysis/README.rst
+   notebooks/Composite_generation/README.rst
+   notebooks/Image_classification/README.rst
+   notebooks/Outputting_data/README.rst
+   notebooks/Workflows/README.rst
+   notebooks/Scripts/README.rst
 
-   notebooks/index
```
andrewdhicks: I'm uncomfortable with including files so deep into a different repo, but I couldn't find a way to get it to work with toctree.

omad: I was trying to do a:
```rst
.. include:: notebooks/index.rst
```
But it doesn't work nicely due to relative paths. We could change that file to specify the path correctly, and then be updated, but, that's a little yucky and wouldn't be viewable directly there.

I also tried doing
```rst
.. toctree::
   :glob:
   :caption: 

   notebooks/*/README.rst
```
But this will include the sections out of order. A possible option would be to rename the folders in the `dea-notebooks` directory.

Thoughts?

omad:
Okay, I've used the globbing feature of `toctree` and have renamed the directories in the `dea-notebooks` repo to keep the chapters in order. See https://github.com/GeoscienceAustralia/dea-notebooks/pull/183

andrewdhicks (approved):
Looks great!

PR #157, opened 2018-08-24
Fix dc config file access for production database

santoshamohan:
## Request for this pull request
Production datacube config file was not correctly retrieved in dea-coherence app.

## Proposed solution
Read datacube config file from the dc environment path and assign that to config_file variable.

- [x] Tests added in NCI test script to test these changes

### DIFF
--- a/digitalearthau/coherence.py
+++ b/digitalearthau/coherence.py
@@ -1,3 +1,4 @@
+import os
 import click
 import structlog
 import collections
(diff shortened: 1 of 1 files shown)


digitalearthau/coherence.py
```diff
@@ -54,7 +55,7 @@ def main(expressions, check_locationless, archive_locationless, check_ancestors,
     """
     global _siblings_count
     uiutil.init_logging()
-    config_file = test_dc_config if test_dc_config else ''
+    config_file = test_dc_config if test_dc_config else os.environ.get('DATACUBE_CONFIG_PATH')
```
omad: This is still potentially fragile. It would be better to set `config_file` to `None` and allow the default configuration discovery process to run.

(or just delete this line and make sure that `test_dc_config` defaults to `None` and use it directly)

omad (changes requested):

digitalearthau/coherence.py
santoshamohan: Updated as per review comments by removing `config_file` variable assignment line and defaulting `test_dc_config` to `None`.

santoshamohan (commented):
Updated

omad (approved):

PR #159, opened 2018-09-11
Index MODIS Ocean Colour Data into ODC Database

omad:
Add a script to enable Product creation and Dataset Indexing
of MODIS Ocean Colour data available on the NCI.

The data is stored in sets of NetCDF file on a daily basis
in `/g/data/u39/public/data/modis/oc-1d-aust.v201508.recent/`
up until 2017, with more to become available (hopefully) soon.

The script `scripts/index_nci_modis_oc.py` can be run in either
with either a `create_product` or `index_data` parameter mode,
and an output directory of OC NetCDF files. It reads the NetCDF
files to create the Product/Dataset definitions, and write them
directly into an ODC database.

It doesn't write out intermediate YAML files, and attempts to create
stable UUIDs for the generate Datasets, based on the file path
and modification time of the underlying NetCDF Data.


andrewdhicks (approved):
This is really good!

PR #160, opened 2018-09-12
Add metadata type gqa_eo, update s2 granule products

uchchwhash:
- Add a new metadata type `gqa_eo`
- Fix Sentinel-2 products (band re-names)

Tested on a datacube instance.


omad (approved):
Thanks Imam!

It's in the DB, we'd better get it in the repo too!

PR #163, opened 2018-09-20
Fix creation time field for Sentinel 2 ARD

jeremyh:
The products coming from wagl have their dataset creation time in `system_information → time_processed`, rather than `creation_dt`. Update the relevant metadata type.

(The field has no indexes so can be updated in-place.)

Note that this metadata type is called `gqa_eo` but is only used by wagl products (and doesn't have any gqa fields?). I'd prefer to rename it to `wagl` too (it has other wagl fields), but that will have to be a different pull request as renaming a metadata type affects more things.


uchchwhash (approved):
:+1:

PR #167, opened 2018-09-28
Remove duplicate nci dea-env test scripts

santoshamohan:
**Reason for pull request**
The `nci` `dea-env` test scripts has been moved to `dea-orchestration` repository. Hence, these scripts can be removed from `digitalearthau` repository.

**Proposed solution**
Remove test script from the following directory:
`digitalearthau/scripts/nci_env_deatest/`


omad (approved):
Thanks

PR #169, opened 2018-10-09
Monitor pbs job status via email

santoshamohan:
`Qsub` job status monitoring via email was only working when a job was `aborted`. Existing code needs to be updated to notify via email whenever a `pbs` job is `successfully executed` or `failed to execute`, in addition to a pbs job being `aborted`.

**Proposed solutions**
- Update the following apps to notify `success`/`failure`/`aborted` status via email:
    -  `dea-submit-ingest` app
    - `dea-submit-sync` app
- Update `run_distributed.sh` script to log additional `pbs` info.

- [x] Tests passed (no update to tests required, as the test scenarios were already written as part of `dea-env` testing)

### DIFF
--- a/digitalearthau/submit/ingest.py
+++ b/digitalearthau/submit/ingest.py
@@ -44,9 +44,9 @@ def list_products():
               type=click.IntRange(1, 48))
 @click.option('--name', help='Job name to use')
 @click.option('--allow-product-changes', help='allow changes to product definition', is_flag=True)
-@click.option('--email_options', '-m', default='a',
+@click.option('--email_options', '-m', default='abe',
               type=click.Choice(['a', 'b', 'e', 'n']),
-              help='Send Email options when execution, \n'
+              help='Send Email when execution is, \n'
               '[aborted | begins | ends | do not send email]')
 @click.option('--email_id', '-M', default='nci.monitor@dea.ga.gov.au',
               help='Email Recipient List')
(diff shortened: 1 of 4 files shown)


digitalearthau/submit/ingest.py
```diff
@@ -44,10 +44,10 @@ def list_products():
               type=click.IntRange(1, 48))
 @click.option('--name', help='Job name to use')
 @click.option('--allow-product-changes', help='allow changes to product definition', is_flag=True)
-@click.option('--email_options', '-m', default='a',
-              type=click.Choice(['a', 'b', 'e', 'n']),
-              help='Send Email options when execution, \n'
-              '[aborted | begins | ends | do not send email]')
+@click.option('--email_options', '-m', default='ae',
```
omad: To stay consistent with the other options, can this please be `--email-options` with a hyphen. The arguments to the function will stay the same, `click` rewrite them.

Can we also default to `abe`, more email spam is a good thing, we can use it to track if and when jobs start.

digitalearthau/submit/ingest.py
```diff
@@ -202,9 +214,15 @@ def stack(queue, project, nodes, walltime, name, product_name, year):
               help='Number of hours to request',
               type=click.IntRange(1, 48))
 @click.option('--name', help='Job name to use')
+@click.option('--email_options', '-m', default='ae',
+              type=click.Choice(['a', 'b', 'e', 'n', 'ae', 'ab', 'be', 'abe']),
+              help='Send Email when execution is, \n'
+              '[a = aborted | b = begins | e = ends | n = do not send email]')
+@click.option('--email_id', '-M', default='nci.monitor@dea.ga.gov.au',
```
omad: Same change for these options with the hyphens.

omad (changes requested):
Sweet. Thanks, couple of tiny changes and we'll merge. :)

omad (approved):

PR #178, opened 2018-10-22
Manage downstream datasets with locationless source dataset

santoshamohan:
# Reason for this pull request
Datasets within the production database had multiple `derived datasets` / `duplicate datasets` derived from the source dataset whose `local_uri` is set to `None`/`Not Available`. This was causing ingestion failure during orchestration process. We need a new tool to manage this scenario and take appropriate action.

**Background**
The above mentioned scenario could arise if we delete the file from the disk and run the `sync` tool with `--update-location` option. Dataset location within the database are set to `None`/`Not Available` by the `sync` tool. So, during `ingest` process, a query from the datacube would result in `index out of range` error and storage file creation fails.

# Proposed solutions
Following are the changes implemented as part of this PR.
1. Fixed `dea-duplicates`
    -  Required changing the setup in `collections.py` for unique fields
    -  Updated the tests since the `CSV` output depends on the unique fields
2. Added `NBART` scenes to `collections.py`
3. Fixed the output directory for the `dea-clean` archived tool
4. Add a new feature to list all the `downstream` datasets whose source dataset has no location
5. Extend `dea-coherence` to check `downstream` datasets
6. Log erroneous datasets to a `CSV` file

- [x] Tests added


digitalearthau/coherence.py
```diff
@@ -1,17 +1,45 @@
...
 from datacube.model import Dataset
 from datacube.ui import click as ui
 from digitalearthau import uiutil
 
-_LOG = structlog.getLogger('archive-locationless')
-_siblings_count = 0
+_LOG = structlog.getLogger('dea-coherence')
+_dataset_count, _siblings_count, _locationless_count, _archive_count = 0, 0, 0, 0
+_downstream_dataset_count, _downstream_summary_prod_count = 0, 0
+_bad_downstream_dataset = collections.deque({})
```
omad: It's not necessary to use a `deque` here, it will work, but so will an ordinary list would be fine.

Using `deque` makes me think that this is used for some special or performance purpose... but it isn't.

Also, it will initialise to empty by default, so the extra `{}` is doubly unnecessary. :)

digitalearthau/coherence.py
```diff
@@ -1,17 +1,45 @@
...
+
+# Product list containing Telemetry, Level1, Level2, NBAR, PQ, PQ_Legacy, WOfS, FC products
+# Level1_Scene datasets are expected not to have any locations (in future)
+PRODUCT_TYPE_LIST = ['ls5_satellite_telemetry_data', 'ls5_level1_scene', 'ls5_nbar_scene', 'ls5_nbart_scene',
```
omad: We should be able to get this list from `digitalearthau.collections`, it'll be too easy for it to fall out of date if we need to list things in more than one place.

digitalearthau/coherence.py
```diff
@@ -1,17 +1,45 @@
...
+                     'wofs_albers']
+
+# Ignore --check-sibling/--check-ancestors for the listed products
+IGNORE_SIBLINGS = ['ls5_nbar_albers', 'ls5_nbart_albers', 'ls5_pq_albers',
+                   'ls7_nbar_albers', 'ls7_nbart_albers', 'ls7_pq_albers',
+                   'ls8_nbar_albers', 'ls8_nbart_albers', 'ls8_pq_albers', 'dsm1sv10']
 
 
-@click.command()
+@click.group(help=__doc__)
```
omad: If there's only the one command, we don't need to have a `click.group`

digitalearthau/coherence.py
```diff
@@ -28,6 +56,10 @@
               is_flag=True,
               default=False,
               help='Check if ancestor datasets have other children of the same type (they may be duplicates)')
+@click.option('--check-downstream-ds',
+              is_flag=True,
+              default=False,
+              help='Check if downstream datasets have their ancestor datasets archived')
```
omad: Make it clear whether this check is for *archived parents* or *locationless parents* or *both*.

digitalearthau/coherence.py
```diff
@@ -111,24 +166,24 @@ def _archive_duplicate_siblings(dc, ids):
...
-    ancestors_archive_count = 0
+    global _siblings_count, _archive_count
+
     dataset = dc.index.datasets.get(dataset.id, include_sources=True)
     if dataset.sources:
         for classifier, source_dataset in dataset.sources.items():
             if source_dataset.is_archived:
                 _LOG.info(
-                    "ancestor.dataset_id",
+                    "ancestor." + str(dataset.type.name) + ".dry_run",
```
omad: Feel free to use the new Python 3.6 style f-strings in `digitalearthau`.
This can be simplified to `f"ancestor.{dataset.type.name}.dry_run"`

digitalearthau/coherence.py
```diff
@@ -139,18 +194,80 @@ def _check_ancestors(check_ancestors: bool,
                     s for s in siblings
                     if s != dataset and not s.is_archived and s.type == dataset.type
                 ]
-                if siblings:
+
+                # Do not process/check siblings for nbar/nbart/pq albers products
+                if siblings and (str(dataset.type.name) not in IGNORE_SIBLINGS):
```
omad: The `(str(dataset.type.name) not in IGNORE_SIBLINGS)` is redundant, it's already checked on line 185.

digitalearthau/coherence.py
```diff
@@ -139,18 +194,80 @@ def _check_ancestors(check_ancestors: bool,
...
+                          dataset_type=str(d_datasets.type.name))
+                _downstream_summary_prod_count += 1
+
+            # Append all the downstream datasets derived from locationless source dataset to the list
+            _bad_downstream_dataset.append(d_datasets)
+
+
+def _fetch_derived_datasets(dataset, dc):
+    to_process = {dataset.id}
+    derived_dataset = collections.deque({})
```
omad: This would be better as `derived_datasets = set()`, and then using `derived_datasets.update(derived)` below will give you a single set of all the derived datasets, instead of the current collection of collections, leading to the nested `for` loops on line 223-4 and 242-3.

digitalearthau/coherence.py
```diff
@@ -139,18 +194,80 @@ def _check_ancestors(check_ancestors: bool,
...
+                          dataset: Dataset):
+    """
+    Need to manage the following two scenarios, with locationless source datasets:
+       1) Identify and list all the downstream datasets (may include level2, NBAR, PQ, Albers, WOfS, FC) that are either
+          Active or Archived.
+       2) Identify and list all the downstream datasets (includes summary products) that are either Active or Archived.
+    """
+
+    # Fetch all the derived datasets using the source dataset.
+    derived_dataset = _fetch_derived_datasets(dataset, dc)
```
omad: It looks like `_fetch_derived_datasets()` is being called twice (here and line 230), but I think it only needs to be called once. It already walks down the entire tree of child datasets.

omad (changes requested):
Hey Santosh, thanks for getting this working.

There's a few things here that it'd be good to clean up before merging. Some are really easy fixes, but some of the logic is also getting hard to reason about and we should sit down and work through it together to see if we can simplify it.

digitalearthau/coherence.py
santoshamohan: Changed `deque` to a simple `set`.

digitalearthau/coherence.py
santoshamohan: Replaced hard-coded values of `product` names to `digitalearthau.collections.registered_collection_names()`

digitalearthau/coherence.py
santoshamohan: Removed `click.group`

digitalearthau/coherence.py
santoshamohan: Modified to python 3.6 style f-strings.

digitalearthau/coherence.py
santoshamohan: Updated to remove any discrepancies.

digitalearthau/coherence.py
santoshamohan: `_fetch_derived_datasets()` is first called with dataset to fetch all derived datasets. Then in the second call, the same function is called with locationless/archived dataset to fetch all the derived datasets of the same.

digitalearthau/coherence.py
santoshamohan: `--check-siblings` command line option removed as this is handled by `--check-downstream` option and by `dea-duplicates` command.

digitalearthau/coherence.py
santoshamohan: Click option `help` string updated.

integration_tests/conftest.py
```diff
@@ -237,6 +255,38 @@ def other_dataset(integration_test_data: Path, test_dataset: DatasetForTests) ->
     )
 
 
+@pytest.fixture
+def ls8_pq_scene_test_dataset(integration_test_data, dea_index) -> DatasetForTests:
+    """A dataset on disk, with corresponding collection"""
```
omad, line 244: What's the purpose of this fixture, and how is it different to `test_dataset`?

integration_tests/conftest.py
```diff
@@ -237,6 +255,38 @@ def other_dataset(integration_test_data: Path, test_dataset: DatasetForTests) ->
...
+    test_data = integration_test_data
+
+    ls8_collection = Collection(
+        name='ls8_scene_test',
```
omad: I think this should be different, this is a PQ dataset...

integration_tests/conftest.py
```diff
@@ -237,6 +255,38 @@ def other_dataset(integration_test_data: Path, test_dataset: DatasetForTests) ->
...
+        query={},
+        file_patterns=[str(test_data.joinpath('LS8*/ga-metadata.yaml'))],
```
omad: This will match more than the newly added PQ dataset.

integration_tests/test_coherence.py
```diff
@@ -0,0 +1,147 @@
...
+    result = CliRunner().invoke(
+        command,
+        args=exe_opts,
+        catch_exceptions=False
+    )
+    assert result.exit_code == 0, "Error for %r. output: %r" % (exe_opts, result.output)
+
+
+def test_check_locationless(ls8_pq_scene_test_dataset: DatasetForTests,
+                            other_dataset: DatasetForTests,
```
omad: It's clear what this test is supposed to be checking, but I don't think the flow is clearly enough explained.

integration_tests/test_coherence.py
```diff
@@ -0,0 +1,147 @@
...
+
+    assert len(ls8_pq_scene_test_dataset.get_index_record().uris) == 0, "Test dataset location should none"
+    assert len(other_dataset.get_index_record().uris) == 0, "Other dataset location should none"
+
+    exe_opts = ['--check-locationless']
+
+    timerange = ["2000-01-01 < time < 2018-12-31"]
+    exe_opts.extend(timerange)
+
+    assert_click_command(coherence.main, exe_opts)
```
omad: This should check for correct output, not only that the coherence tool ran without errors

integration_tests/test_full_sync.py
```diff
@@ -26,7 +26,8 @@
 
 def test_new_and_old_on_disk(test_dataset: DatasetForTests,
                              integration_test_data: Path,
-                             other_dataset: DatasetForTests):
+                             other_dataset: DatasetForTests,
+                             ls8_pq_scene_test_dataset: DatasetForTests):
```
omad: I don't think that `ls8_pq_scene_test_dataset` should be added to these tests. Maybe it was getting pulled in due to the fixtures not setting up the `Collections` correctly, as I've pointed out above.

omad (dismissed):
Hi Santosh, I think this is looking much better. I think there's still a few issues with the changes in tests.

Also, please try to list all the changes in a PR, for example for this PR:
-   Fixed `dea-duplicates`
    -   Required changing the setup in `collections.py` for unique fields
    -   Updated the **tests** since the CSV output depends on the unique fields
-   Added **NBART** scenes to `collections.py`
-   Fixed the output directory for the `dea-clean archived` tool
-   Extend `dea-coherence` to check downstream datasets

integration_tests/conftest.py
santoshamohan: `test_dataset` contains `level1` dataset that would be skipped by the `coherence` tool. Hence, I had to add new `pq_scene` fixture.

integration_tests/conftest.py
santoshamohan: Agreed. Updated the collection name to be `ls8_pq_scene_test`.

integration_tests/conftest.py
santoshamohan: Updated the directory name to match only the newly added `PQ` dataset.

integration_tests/test_coherence.py
santoshamohan: Added more comments to aid the test flow.

integration_tests/test_full_sync.py
santoshamohan: Reverted back the changes and updated `conftest.py` to pull correct `collections`.

integration_tests/test_coherence.py
santoshamohan: Updated test to check the expected output.

PR #187, opened 2018-11-18
Configure travis to push built HTML to our s3 bucket

tom-butler:
Use s3 and cloudfront so we can have more control over the hosted docs.
Infrastructure already exists
travis used has the permissions required to write files to this bucket


omad (approved):
Noice one.

PR #188, opened 2018-11-23
More coherence check improvements

omad:
Hi Santosh, can you please have a look at this refactoring I've attempted on the coherence tool, and make sure I haven't broken or messed anything up. I was having trouble understanding what it was doing, and hope that this makes things a little clearer.

I think that there's still room for improvement. Possibly by removing some of the optional command line arguments and hard coding the behaviour. There's too many different ways that the tool can operate, and I don't think it's all that useful.


santoshamohan (approved):
Thanks for refactoring the coherence code, Damien.
The code now is more readable and maintainable.

PR #195, opened 2018-12-04
WOfS wet/dry seasonal summaries

bellemae:


digitalearthau/config/products/wofs_apr_oct_summary.yaml
```diff
@@ -0,0 +1,37 @@
...
+metadata_type: eo
+
+metadata:
+  product_type: wofs_apr_oct_summary
+  format:
+    name: 'NetCDF'
+  instrument:
+    name: 'TM,ETM,OLI'
+  platform:
+    code: 'LANDSAT_5,LANDSAT_7,LANDSAT_8'
```
andrewdhicks, line 12: The `instrument` and `platform` sections probably aren't needed in the `metadata` section if they are a mix (as it is really a `dataset matching` section).  
Not really an issue until Landsat 9 is launched, but we can always code to make it understand lists of platforms, etc.

andrewdhicks (approved):
:+1:

PR #198, opened 2019-01-08
travis fix

dunkgray:
Hi Damien,
Travis was failing on the develop branch.  This is a fix.

Cheers
Duncan


Kirill888 (approved):
This will fix current failure, but we should really use frozen set with `requirements-test.txt` file like we do in opendatacube repo.

PR #199, opened 2019-01-09
Deadm 1007

dunkgray:
Modifications to get all available daily MODIS ocean colour data indexed into DEA.


omad (approved):
Looks great, thanks Duncan.

I'm going to say it's safe to merge all these commits into a single commit, since there's a lot of unnecessary churn.

The main explanation from commit messages I can see is that the month of January 2018 has been excluded because it includes some bad data.

PR #203, opened 2019-01-24
fix metadata-type again

uchchwhash:
Turns out that just one metadata type won't do.

The Sentinel-2 ARD products in the current database have their metadata type set to `gqa_eo` whose GQA fields are not yet searchable. This PR fixes that, with the correction over the last one that `gqa_final_qa_count` should actually be `gqa_final_gcp_count` to conform with the existing datasets.

The upgraded collection will however call that field `gqa_final_qa_count` which necessitates another version of the EO metadata type (called `eo_plus` in this PR).


jeremyh (approved):
Thanks Imam 👍

PR #207, opened 2019-03-15
Fix build

Kirill888:
This is mostly changes from @omad, but we need them now in develop.

looks like newer netcdf4 libs break some stuff when running on Travis at least, so I'm fixing NetCDF4 version for now.


ashoka1234:
Thanks Kirill. All good now.

dunkgray (approved):

Issue #208, opened 2019-03-15
Creating sample netcdf stops working with netCDF4>1.4.1


Kirill888:
Something changed in the way string/byte data is handled by netCDF4 lib starting from version 1.4.2 and up. Last version where this code works is 1.4.1:

https://github.com/GeoscienceAustralia/digitalearthau/blob/e21f51bc809d2b3c374de311493cf730d67ec1c8/integration_tests/test_move.py#L217-L220

Kirill888:
Issue turned out to be with binary distribution of `netCDF4` library, calling `import neCDF4` before calling `import numpy` makes code above work. So does compiling `netCDF4` locally instead of pulling in binaries. So it's not API changes or anything, just bad binary distribution.

Kirill888:
Won't fix, problem is with binary distribution upstream

PR #209, opened 2019-03-15
Generate sample netcdfs in a style closer to production

Kirill888:
This started as investigation of #208, but problem in that issue is due to binary package distributed by `netCDF4`, it has some binary incompatibility with `numpy`, problem goes away if latest `netCDF4` package is compiled on the system rather than pulling in binaries. I'm not fixing that instead just pinning `netCDF4` to the lastest version that works, which is `1.4.1`.

Changes are:

1. Create time axis that contain time
2. Use `datacube.drivers.netcdf.writer.*` helper methods rather than `netCDF4` lib directly
3. Using `netCDF4==1.4.1` in travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ before_install:
   - export GDAL_DATA="$(gdal-config --datadir)"
   - pip install --upgrade pip
   - pip install --progress-bar off coveralls codecov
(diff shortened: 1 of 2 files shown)


omad (approved):
Thanks

PR #217, opened 2019-04-18
Jupyterlab extensions

Kirill888:
Adding 3 more lab extensions:

- git: stage commit from jupyterlab 
- nbdime: visual diff of notebooks
- ipywidgets for jupyterlab: work with ipywidgets in jupyterlab

I have tried these extensions with manual install and that worked fine, but not sure how this is supposed to be tested.

### DIFF
--- a/modules/dea-env/environment.yaml
+++ b/modules/dea-env/environment.yaml
@@ -97,6 +97,7 @@ dependencies:
   - gunicorn
   - hdmedians
   - ipython-sql
+  - jupyterlab-git
   - jupyterlab_github
   - meinheld
   - modernize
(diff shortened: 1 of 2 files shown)


omad:
Awesome Kirill. I'll get @santoshamohan to migrate the changes to dea-orchestration. That's where we have been maintaining the environment creation scripts.

Need to clean up the current duplication, and maybe move it back here, since I think it makes more sense in this repo.

omad:
NCI DEA module is managed over at [GeoscienceAustralia/dea-orchestration](https://github.com/GeoscienceAustralia/dea-orchestration) repo now. I think most of these changes have probably gone in already, or are out of date by now.

PR #219, opened 2019-04-30
ASTER Indexing Improvements

omad:
Hi Ashoka,

I'm going to leave some comments on your pull request too. These are just some changes I made while going through the code.

In summary, the changes are:
- Remove the gigantic 90mb HDF file used for testing, and replace it with a shrunken version containing the same metadata. (We need to be careful when doing the final merge that we don't accidentally include the big file in the repo history, and bloat it unnecessarily).
- Get the tests working in my environment
- Documentation cleaning and editing
- Bits of other code cleaning


ashoka1234:
Looks pretty good. Thanks

PR #222, opened 2019-05-16
Update structlog dependency

jeremyh:
Structlog changed the meaning of their `[dev]` package: it now installs test+doc dependencies that we don't need.

Our build broke when they did this, and we fixed it in two separate branches by adding Colorama (once at the top of the dependencies and once at the bottom), which were later merged together.

This removes `[dev]` from structlog and removes the duplicate Colorama.

See: hynek/structlog@d6b7da7

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,19 +37,17 @@ setup(
         'pytest-runner'
     ],
     install_requires=[
-        'colorama',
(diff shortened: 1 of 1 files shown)


uchchwhash (approved):
Build still failling. Should not prevent me from approving though. :smile:

PR #228, opened 2019-07-30
Product definitions for triple MAD products

bellemae:


omad (approved):
Brilliant!

Thanks Belle.

PR #230, opened 2019-08-02
Fix docs build in Travis CI

omad:
Travis CI `develop` branch build is still broken. I think this is because of an old `jupyter_sphinx` package the docs are looking for but not actually using.

The documentation builds correctly on my local machine.

This PR:
- removes the `jupyter_sphinx` package from `docs/conf.py`.
- adds an opt in [pre-commit](https://pre-commit.com) configuration


jeremyh (approved):
👍

PR #231, opened 2019-08-08
Import & update the collection 3 product definitions

jeremyh:
Imported from @Kirill888 's originals in [dea-proto](https://github.com/opendatacube/dea-proto) with git history intact.

Made changes based on the feedback register:

- Added simple aliases for red/green/blue etc for nbart.
- Expanded the product descriptions (A little. Maybe people will want more?)
- There's now multiple metadata types (and a minimal default 'eo3') so that products wont have a large number of confusing null fields. (original feedback was for Level1, but it will affect all non-ARD products)
- Added aliases without `oa_` prefixes
- Fixed invalid nodata values (uint8 -> -999)
- Added the new naming conventions for products `*.odc-product.yaml`

Misc changes
- Removed unnecessary indexes (for platform/instrument/etc), which don't differ within a product.
- Ran a ~ruamel whitespace~ headless pycharm autoformat on the whole history.

Still outstanding:
- ~nodata value review~ (✓done)
- a second/fmask cloud field (needs to be pushed upstream to the datasets first if we're doing it)


digitalearthau/config/eo3/products/ard_ls5.yaml
```diff
@@ -0,0 +1,177 @@
...
+    dtype: float32
+    nodata: .nan
+    units: '1'
+-   name: oa_azimuthal_incident
+    dtype: float32
+    nodata: .nan
+    units: '1'
+-   name: oa_combined_terrain_shadow
+    dtype: uint8
+    nodata: -999
```
Kirill888: this one needs to be fixed as raised by @simonaoliver in other products too

digitalearthau/config/eo3/products/ard_ls7.yaml
```diff
@@ -0,0 +1,190 @@
...
+    dtype: float32
+    nodata: .nan
+    units: '1'
+-   name: oa_azimuthal_incident
+    dtype: float32
+    nodata: .nan
+    units: '1'
+-   name: oa_combined_terrain_shadow
+    dtype: uint8
+    nodata: -999
```
Kirill888: nodata value outside of dtype range

digitalearthau/config/eo3/products/ard_ls8.yaml
```diff
@@ -0,0 +1,203 @@
...
+    dtype: float32
+    nodata: .nan
+    units: '1'
+-   name: oa_azimuthal_incident
+    dtype: float32
+    nodata: .nan
+    units: '1'
+-   name: oa_combined_terrain_shadow
+    dtype: uint8
+    nodata: -999
```
Kirill888: and again

digitalearthau/config/eo3/metadata_eo3.yaml
```diff
@@ -0,0 +1,85 @@
+---
+name: eo3
+description: Default EO3 with no custom fields
+dataset:
+    id: [id]  # needs to be fixed
```
Kirill888: I know I put this comment in myself `# needs to be fixed` but it can be misinterpreted as "needs fixing" while it really means "has to be this value and not any other"

digitalearthau/config/eo3/products/ard_ls5.yaml
jeremyh: :+1:

digitalearthau/config/eo3/products/ard_ls7.yaml
jeremyh: :+1:

digitalearthau/config/eo3/products/ard_ls8.yaml
jeremyh: :+1:

digitalearthau/config/eo3/metadata_eo3.yaml
jeremyh: Updated the comment

jeremyh:
Products now validate against the sample data:

```
$ eo3-validate --thorough /g/data/v10/agdc/jez/digitalearthau/digitalearthau/config/eo3/products/*.yaml `fd odc-metadata.yaml`
✓ ard_ls5
✓ ard_ls7
✓ ard_ls8
✓ l1_ls5
✓ l1_ls7
✓ l1_ls8
✓ ga_ls5t_ard_3-0-0_092084_2009-12-17_final.odc-metadata
	I different_nodata oa_combined_terrain_shadow nodata: product 255 != dataset None
✓ ga_ls7e_ard_3-0-0_092084_2016-08-23_final.odc-metadata
	I different_nodata oa_combined_terrain_shadow nodata: product 255 != dataset None
✓ ga_ls8c_ard_3-0-0_092084_2016-06-28_final.odc-metadata
	I different_nodata oa_combined_terrain_shadow nodata: product 255 != dataset None

valid: 3 infos
```

The `combined_terrain_shadow` dataset nodata is None according to @sixy6e as it never has any nodata values, and 255 will be used in the product as the ODC filll value. 

(Because of the fill-value/different usage of these fields, I've changed the validator to allow nodata values to differ, only printing an info)

jeremyh:
(in keeping the Git history of this branch clean I've made this issue history a mess :slightly_smiling_face: )

Kirill888 (approved):
:+1:

PR #237, opened 2019-08-30
Remove the docs, point to new location

### LINKED ISSUE #232: Have one and only one URL for documentation
We currently build the DEA user guide from this repository. Identical copies are published to *two* different URLs. This build and deploy runs in TravisCI.

1) https://geoscienceaustralia.github.io/digitalearthau/ (Served in GitHub Pages from the `gh-pages` branch of this repository)
2) https://docs.dea.ga.gov.au/ (Served from AWS S3+Cloudfront)

We should simplify to only use the second url, and delete the publishing to GitHub Pages. It would be nice to set up redirects so that any old links don't immediately break.

AFAICT, the best way to set up redirects is to replace each `.html` file in the `gh-pages` branch with content similar to, but including paths to and from each html file:

```html
<!DOCTYPE html>
<meta charset="utf-8">
<title>Redirecting to https://docs.dea.ga.gov.au/</title>
<meta http-equiv="refresh" content="0; URL=https://docs.dea.ga.gov.au/">
<link rel="canonical" href="https://docs.dea.ga.gov.au/">
```

I've sent a request to get the link at http://www.ga.gov.au/dea/news updated to the new URL. Once this happens, I can go ahead and make the change, assuming no-one has any other objections.

omad:
There's a new repository for the DEA User Guide at https://github.com/GeoscienceAustralia/dea-docs/ . This PR removes the docs from this repo, and points to their new location.

This new location is to make it easier for people from science teams and Frontier SI to maintain and contribute to the documentation. We can manage permissions and branch restrictions differently, and it also simplifies the CI setup.

The new repo is already publishing to https://docs.dea.ga.gov.au/, so this PR removes that deployment.

I'm in the process of sorting out redirects from the `gh-pages` hosted docs, see #232.

Closes #232


omad:
This is ready for merging now.

I've replaced the `gh-pages` branch of this repository with HTML redirects, and tested that they work. This should *significantly* reduce the size of cloning this repo.

We now need to stop building or pushing the docs at all from this repo.

alexgleith (approved):
Love PRs that remove stuff!

PR #238, opened 2019-09-04
Temporary configuration file change to keep an old production job happy

omad:
Can be reverted in a week or two when that system is decommissioned.


tom-butler (approved):
Thanks!

PR #246, opened 2019-12-19
Delete files that are wrong. 

abhijeetkumarga:
Accidentally pushed incorrect files.


omad (approved):
We can do this all day! 🤣

PR #252, opened 2020-02-10
Let storage settings pass through in qsub

uchchwhash:
- Enable qsub to pass through the `storage` setting that is now required by Gadi
- Change default `mem` setting to `medium` (4G)


santoshamohan (approved):
+1

PR #261, opened 2020-06-26
Product definition updates for C3

jeremyh:
(see commit messages below: they're clean, so no point repeating here?)

The only potentially-breaking change is increasing the ODC requirement to version 1.8, as its needed for license fields in products. But all DEA environments should now use 1.8, I assume?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,6 @@ before_install:
 install:
   - travis_retry pip install --upgrade pytest
   - travis_retry pip install codecov cython boto3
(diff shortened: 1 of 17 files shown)


jeremyh:
@simonaoliver: Are these licenses correct for our Landsat products?  `license: CC-BY-4.0` (ie: https://creativecommons.org/licenses/by/4.0/ )

simonaoliver:
Yes - I believe CC-BY 4.0 applies here @jeremyh .

simonaoliver (approved):
description now agrees with convention

PR #262, opened 2020-07-21
Add Landsat C3 NBART Product Definitions

alexgleith:
These three product definitions are required when we upload data to AWS.

The major change is removal of the NBAR bands. And some suggestions are product and file names.


jeremyh (approved):
:+1: Looks good.

Ideally we'll add ` license: CC-BY-4.0` too. That was added to the normal products in #261, but has been sitting open, blocked by the same Travis failure as this one (a sudden ODC install issue I've been meaning to look into)

alexgleith:
> Ideally we'll add ` license: CC-BY-4.0` too. 

It probably should be in a different PR, since it's not in the C3 ARD product definitions yet either...

jeremyh:
Woo-hoo, success :) 

Feel free to click merge if it looks ok.

PR #271, opened 2020-09-24
Delete config directory

alexgleith:
This config directory is being replaced by the one that's being created in this PR: https://github.com/GeoscienceAustralia/dea-config/pull/593

Todo: we need to work out where to put the old `config/ingestion` and `config/products` folders...


jeremyh:
There's a few places in the code that use these, which is why the tests are failing. They're also bundled into the pip install of the package.

eg.
- The `dea system init` command which creates a datacube with DEA's config and products (see [here](https://github.com/GeoscienceAustralia/digitalearthau#creating-a-dea-configured-data-cube)).
- The pytest plugin, which adds products too (used by projects like Explorer to set up / tear down ODC test instances and to test with real DEA products)

We maybe need an import, or git submodule or something, if the plan is to move them there.

jeremyh:
I also like that this repo lints them and tests that they can be added to datacube... we've had issues with invalid yamls before.

alexgleith:
Ok, well how about I set up a submodule and implement linting over in `dea-config. I'll give it a whirl and come back.

kieranricardo:
@alexgleith @jeremyh are there still places in the code that use the `dea-config` repo? I did a quick search through the `develop` branch here and couldn't find any mention of it

alexgleith:
I think if we get the CI working we'll be ok.

CI working means either ignoring linting in the submodule, or fixing the linting in the other repo!

kieranricardo:
I feel like it'd be nicer to get the linting to work! `yamllint` doesn't seem to have any auto-formatting which is annoying, but there's some others out there that do like `ruamel.yaml.cmd`. 

`ruamel.yaml.cmd` formatting is inconsistent with `yamllint` but one option is to replace the CI tests with `ruamel.yaml.cmd` and then use its auto formatting to fix up this PR. What do you think @alexgleith?

alexgleith:
Yaml linting example is over here: https://github.com/digitalearthafrica/config/blob/master/.github/workflows/linting-yaml.yml

I use the redhat linter for VS Code and it works... 

But I trust you to sort this out however you want to!

image

kieranricardo:
thanks! I just wanted to check that we aren't wedded to `yamllint` for some reason :)

 I have `yamllint` working but my problem is that doesn't format files, it only lints them (and I don't wanna do everything manually!). And I couldn't seem to find a formatter that kept `yamllint` happy

dunkgray:
Where is are current files like;
digitalearthau/config/eo3/products/ard_ls5.odc-product.yaml
being moved to?

Issue #277, opened 2021-02-24
The NIR band has an alias of red


alexgleith:
https://github.com/GeoscienceAustralia/digitalearthau/blob/085d69cb6ecc7f277b37be80531de01208ec9caf/digitalearthau/config/eo3/products-aws/ard_ls7.odc-product.yaml#L44

https://github.com/GeoscienceAustralia/digitalearthau/blob/085d69cb6ecc7f277b37be80531de01208ec9caf/digitalearthau/config/eo3/products/ard_ls7.odc-product.yaml#L44

These need to be fixed and then applied on 3 AWS DBs and the NCI DB

PR #278, opened 2021-02-25
Fix nir-band alias in ARD C3 products

jeremyh:
Incorrectly marked as 'red', as reported in #277


emmaai:
can it be approved? @alexgleith @simonaoliver

alexgleith (approved):
Approved. I'm happy to apply I'm all the environments later today.

PR #283, opened 2021-03-17
add water_observed bit definition

emmaai:
As title. The bit `water_observed` exists in `c2`, which gives the flexibility to make mask. Would like it to be in `c3` to keep consistency.


alexgleith (approved):
Norman approved on Slack 19/3/2021

PR #286, opened 2021-04-20
Adding LS8 NBART Geomedian product definition

Kirill888:


alexgleith:
Although we're not linting, this should be named with `.yaml` and technically should have three `---` at the top, for some reason.

I'm happy with just naming it `yaml` please :-)

alexgleith (approved):

PR #295, opened 2021-05-20
created new eo3 type

kieranricardo:
Adds a new metadata type that contains the standard eo3 fields plus the searchable fields here: https://explorer.sandbox.dea.ga.gov.au/products/s2a_ard_granule


alexgleith:
Can we not just use the Landsat one? https://github.com/GeoscienceAustralia/digitalearthau/blob/develop/digitalearthau/config/eo3/eo3_landsat_ard.odc-type.yaml

kieranricardo:
the Landsat one has a bunch of extra fields - is that important? Do searchable fields have to exist in dataset documents?

alexgleith:
Nope, extra fields is fine. No problem with that...

Let's re-use it :-)

kieranricardo:
sounds good :)

alexgleith:
I'm happy if @omad is happy!

omad (approved):
LGTM!

I'm sad about the poor naming, but I'll get over it and we can consider fixing it some other time. :)

PR #305, opened 2021-08-24
Build + CI fix and improvement

jeremyh:
In short:
- Remove old code that is broken and unused.
- Reduce the required dependencies heavily (niche dependencies now optional).
- Run `eo3-validate` automatically on the product definitions, and fix a few warnings.
- Move CI to use Python 3.8, as with ODC core repo.
   (the alternative is to debug which dependencies are currently broken on 3.6. But if ODC only tests with 3.8, why not this repo?)

----
In Detail:
- [x] Remove the DEA Stacker (it's breaking the build)

ODC 1.8.4 [removed the datacube_apps module](https://github.com/opendatacube/datacube-core/blob/develop/docs/about/whats_new.rst#v184-6-august-2021), so the stacker code fails to import, and breaks our build.

We don't stack now, so let's delete it.

- [x] Remove pds tools install. None of our code uses it anymore.
- [x] Remove the test that verifies the exact list of products. This is a hassle to update all the time, and yamllint + eo3-validate give avoid the problems this was added for.
- [x] Run eo3-validate on products automatically. Catches common errors.
- [x] Add missing license to geomedian, [based on CMI](https://cmi.ga.gov.au/data-products/dea/140/dea-surface-reflectance-geomedian-landsat#license) (validator complained about this)
- [x] Remove redundant aliases (when they're the same name as the band)
- [x] **Move the CI to use Python 3.8**. This may be controversial, but `pip install` is repeatedly freezing on Python 3.6 now, and in investigating I noticed ODC core repo itself has moved to 3.8 (despite still officially supporting 3.6?).
- [x] Replace the custom tuple serialisation code with [cattrs](https://github.com/Tinche/cattrs#cattrs) (since it was broken on newer Python, and cattrs is simpler. The tests were very helpful here.)


jeremyh:
Github CI seems to be hanging indefinitely on the pip install step. Repeatedly.

uchchwhash (approved):
💯

PR #306, opened 2021-08-25
Update provisional products+examples

jeremyh:
(split these out of #305 as that's a bigger change: the build wont work until that is merged, but I've tested this manually.)

The new examples no longer have nbar bands.

Move them to the AWS products folder, alongside our other non-NBAR ARDs.


uchchwhash (approved):
Thanks @jeremyh!

PR #310, opened 2022-02-09
adding search for landsat_product_id

dunkgray:
Hi,
As Dave suggested in slack, here's the pull request to add landsat_product_id  in the searchable fields in Landsat c3 ard NCI.


omad (approved):

omad:
Looks good Duncan. I'll put these up on the NCI.

Can you please update the file in `dea-config` as well before we update the AWS databases. We just checked and they are identical according to the different ODC Explorer instances.

https://github.com/GeoscienceAustralia/dea-config/blob/master/product_metadata/eo3_landsat_ard.odc-type.yaml

PR #316, opened 2022-08-16
LS9 draft products

jeremyh:
... we still haven't moved everything to dea-config (and it doesn't seem to have a space for validation dataset examplesat the moment). So for the moment I'm putting it in a branch here


dunkgray (approved):
looks good. 
Note the provisional products are also defined here;
https://github.com/GeoscienceAustralia/dea-config/tree/master/products/baseline_satellite_data/provisional

PR #317, opened 2022-10-19
update env yaml

Ariana-B:
Remove extraneous dependencies from dea-env environment.yaml, update to python 3.8

Tested on gadi


omad (approved):
:+1:

PR #318, opened 2022-10-19
Update dea env

Ariana-B:
Update modulespecs to use python 3.8 and mambaforge


omad (approved):
Lets see how it goes (and what breaks next!)

PR #319, opened 2022-10-19
Update dea env

Ariana-B:
Add nodejs to dependencies as required by final jupyter installations


omad (approved):
Looks good. No `nodejs` is definitely causing the `jupyter labextension install` failure.

PR #325, opened 2022-11-14
Update factory index creation 

Ariana-B:
Datacube Index should not be initialised directly and is breaking code. Changed to use `index_connect()` instead.


omad (approved):
🤩

